PR #42, opened 2025-01-30
Update frontend dependencies

kyasbal:
TODOs:
- [x] Update Node.js version
- [x] Update package.json
- [x] Update angular.json
- [x] Migrate non standalone components to standalone components
- [x] Remove ngx-env and replace it with the define(https://angular.dev/tools/cli/build-system-migration#build-time-value-replacement-define)
- [x] Update style related errors
- [x] Fix broken frontend tests
- [x] Reduce dependencies from modules that become removable after migrating components to standalone

Major notable changes:

* Angular 17 -> 19
* Migrate components to standalone as much as possible
* Get rid of ngx-env


kyasbal:
Now it works with Angular 19. I need to fix tests and reduce non necessary module dependencies.

image

web/src/app/header/titlebar.component.html
```diff
@@ -15,11 +15,11 @@
 -->
 
 <h1 class="title">Kubernetes History Inspector</h1>
-<p class="version">v {{ ("NG_APP_VERSION" | env) || "Dev" }}</p>
-<p class="viewer-mode" *ngIf="'NG_APP_VIEWER_MODE' | env">
+<p class="version">{{ version }}</p>
+<!-- <p class="viewer-mode" *ngIf="'NG_APP_VIEWER_MODE' | env">
```
renamoo, line 19: Is it intended to keep this?

web/src/environments/environment.prod.ts
```diff
@@ -19,6 +19,7 @@ import { PublicKHIExtension } from 'src/app/extensions/public/module';
 export const environment = {
   production: true,
   viewerMode: false,
+  apiBaseUrl: '',
```
renamoo, line 22: Just want to check if I understand it right. 
In dev profile, `this.baseUrl` = `http://localhost:8080` + " ".  
In prod, `this.baseUrl` = " " + `http://localhost:XXXX (host currently used)` ?

renamoo (commented):
Let me check two points

web/src/environments/environment.prod.ts
kyasbal: I've added a comment on code to clarify this.
https://github.com/GoogleCloudPlatform/khi/pull/42/commits/bbdef50f272ebeb24f3873197ed3a52da01ca16b#diff-56a917cf4582773a0908e0f38e0e5aa50232d80507c5fdcf861e29b381fb3051R22

Your understanding is right, and this configuration varies to because of environment setup for development.

web/src/app/header/titlebar.component.html
kyasbal: I created another issue #46.
I need to add the API to get the viewer mode status from it instead of env variable on the build time.

renamoo (approved):
Thank you for the responses! LGTM

kyasbal:
I'll merge this next week not to break frontend code on the main branch when I can't fix it soon. (I'm on a travel this weekend)

PR #52, opened 2025-02-03
Add development guide

kyasbal:
* Add instructions to start development
* Add `jq` command on the prerequisite list
* Remove `Google Chrome` from the prerequisite in Japanese README to match with English one

### DIFF
--- a/docs/en/development-guide.md
+++ b/docs/en/development-guide.md
@@ -2,9 +2,18 @@
 
 ## Run your first build
 
-Follow [the "Run from source code" section](../../README.md) on README.
+Follow [the "Run from source code" section](../../README.md#run-from-source-code) on README.
 
 ## Setup environment for development
+
+### Fork KHI repository
+
+You can't create a new branch our repository directly. Please fork our repository on your account to modify.
+
+### Setup commit signature verification
+
+Please check [this document](https://docs.github.com/en/authentication/managing-commit-signature-verification) to make sure your commits are signed.
+Our repository can't accept unsigned commits.
 
 ### Setup Git hook
 
@@ -49,7 +58,7 @@ Save the following code as `.vscode/launch.json`.
 }
 ```
 
-You can run KHI with VSCode and features like break points are available with it.
+You can run the server with VSCode. You can refer [this document](https://code.visualstudio.com/docs/languages/go) for more details.
 
 ### Run frontend server for development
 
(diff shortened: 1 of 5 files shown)


docs/en/development-guide.md
```diff
@@ -0,0 +1,82 @@
...
+
+This will build the frontend code with [the configuration to access APIs on `localhost:8080`](../../web/src/environments/environment.dev.ts).
+You can use KHI with accessing `localhost:4200` instead of `localhost:8080`. Angular dev server automatically build and serve the new build when you change the frontend code.
+
+### Run test
+
+Run the following code to verify frontend and backend codes.
+
+```shell
+$ make test
```
renamoo, line 70: It does not work unless you have Angular CLI installed or we change 
```
test-web: prepare-frontend
	cd web && ng test --watch=false
```
to `npx ng ...`

docs/en/development-guide.md
```diff
@@ -0,0 +1,82 @@
...
+                "maxStructFields": -1
+            },
+        }
+    ], 
+}
+```
+
+You can run KHI with VSCode and features like break points are available with it.
+
+### Run frontend server for development
```
renamoo, line 54: Just checking, do we want to add Run backend section? Or, since we mention "You can run KHI with VSCode and features like break points are available with it." in the VS Code section, we consider it's obvious and skip that part?

docs/en/development-guide.md
```diff
@@ -0,0 +1,82 @@
+# Development guide
+
+## Run your first build
```
renamoo, line 3: Just checking, do we want to mention a process to fork repo and set up a signature for commit somewhere? 
We might be able to skip that if that's very common for OSS but I don't have much experience so not sure about that.

docs/en/development-guide.md
```diff
@@ -0,0 +1,82 @@
...
+
+Follow [the "Run from source code" section](../../README.md) on README.
```
renamoo, line 5: Maybe we can use `../../README.md#run-from-source-code`?

docs/en/development-guide.md
kyasbal: I changed the make command to use npx. Thank you for catching this!

docs/en/development-guide.md
kyasbal: We should cover the process. I added sections.

docs/en/development-guide.md
kyasbal: Fixed

renamoo (approved):
Thank you for the updates. LGTM

PR #59, opened 2025-02-04
Refactor & bug fixes on node log parser

kyasbal:
## Background

This PR includes changes for the features below.

### ID replacement feature

The node log parser used in KHI finds container ID or pod sandbox ID in log bodies. This helps us to understand containerd/kubelet behavior well with KHI parsing these IDs from the log like below.

```
time="2024-11-29T13:06:01.547380841Z" level=info msg="StopPodSandbox for \"c6a05fb78aff6c14288a71269955e859c802385f9493657ca3234d426a07e5fc\""
```

KHI convert this log message to the following message in the summary view with replacing pod ID with the actual pod names.

```
StopPodSandbox for "c6a05fb...(1-5-workloads/init-failure-deployment-585755d8c4-z9kj7)"
```

### Severity parser

Node logs doesn't always contain the `.severity` field on these logs.

```
E1129 13:12:22.699288    2062 remote_runtime.go:432] "ContainerStatus from runtime service failed" err="rpc error: code = NotFound desc = an error occurred when try to find container \"e8b11f6ab29fad0f64d5af22efcc836b915359adabd8b4a25b4bae6eedaee079\": not found" containerID="e8b11f6ab29fad0f64d5af22efcc836b915359adabd8b4a25b4bae6eedaee079"
```

KHI parse the message and reflect the severity indicated in the log. In this time, it's error log because the first character is `E`.

## Changes

* Refactor severity parsers and fixed bug not reflecting the right severity.

* Changed algorithm to associate Ids with logs. 
Previously, we had added condition branches for every types of logs. From this change, parser seeks if any of Ids found on the node so far is included. This is a bit computational heavy compared to the past implementation, but it's robust implementation. We'll optimize it later if this implementation is heavily affecting parsing time. But so far, I don't observe such behavior.


pkg/parser/k8s/klog.go
```diff
@@ -18,15 +18,31 @@ import (
...
 
-	"github.com/GoogleCloudPlatform/khi/pkg/log/schema"
+	"github.com/GoogleCloudPlatform/khi/pkg/model/enum"
 )
 
-var KLogSeverityFieldAlias = "@severity"
+// severityStringNotation maps string notation of severity found in logs to the severity types used in KHI.
+var severityStringNotation = map[string]enum.Severity{
+	"INFO":    enum.SeverityInfo,
+	"info":    enum.SeverityInfo,
```
RyuSA, line 27: FYI: the lower letter(level=...) may be rely on here.
https://github.com/sirupsen/logrus/blob/v1.9.3/logrus.go#L59

ref: https://github.com/containerd/containerd/blob/v2.0.2/vendor/github.com/containerd/log/context.go#L71

RyuSA (approved):
LGTM(added small memos)

pkg/parser/k8s/klog.go
kyasbal: Then I'll add `panic` to be the fatal level on KHI. `trace` and `debug` to be default

RyuSA (approved):
LGTM

Issue #54, opened 2025-02-03
Setup .devcontainer


kyasbal:
Create devcontainer configurations to make preparation for development easier.

Sozhan308:
Hello @kyasbal,

Thank you for open-sourcing the project! I’d love to contribute to the KHI project. If this task is still available, I’d be happy to take it on. Let me know if it works for you or if there’s anything I should know.

Thanks!

RyuSA:
Hi @Sozhan308 , thank you for your interesting! 
I'm looking forward to your contribution :) 

To develop and build this project, you will need Node.js, Go, gcloud, etc. Please refer to the Requirements in the README for details!
https://github.com/GoogleCloudPlatform/khi#run-from-source-code

Note: `gcloud` archive is available on this GCS
https://console.cloud.google.com/storage/browser/cloud-sdk-release

Sozhan308:
Hi @RyuSA,

Thank you for sharing the details! I’ve reviewed the README and the requirements, and I’ll make sure to set up Node.js, Go, and gcloud as specified.

I’ll get everything set up and If I have any questions along the way, I’ll reach out. Thanks again for your guidance!

kyasbal:
@Sozhan308 Thank you for being interested in this!

Before starting this issue and commit something, you would need to make sure configuring commit signature verification.
https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

This repo can't merge any commits without signatures, and it's a bit tricky to add signatures after creating commits. 
Feel free to ask me anything if you trouble with getting it done.

kyasbal:
Fixed by #62

PR #62, opened 2025-02-06
feat(dev): add development container configuration (#54)

Sozhan308:
# Add Development Container Configuration (#54)
## Overview
This PR introduces devcontainer setup configurations. The configuration includes necessary tools, dependencies, and post-setup scripts for Go and Node.js development.

### Added Development Container Config Files
- Dockerfile : Sets up Ubuntu-based development environment with:
  
  - Go development tools
  - Node.js and npm
  - Google Cloud SDK and kubectl
  - User permissions and workspace setup
  
- docker-compose.yml : Configures:
  
  - Volume mounts for source code and caches
  - Build arguments and environment variables
  - Docker socket access for container operations
  - Network and security settings

- setup-dev.sh : Post-creation setup script that:
  
  - Verifies required tool installations
  - Sets up Go development environment
  - Configures npm and Angular CLI
  - Sets Git configurations

### Testing

Verified the following tools inside the devcontainer

```bash
go version
node --version
npm --version
ng --version
gcloud --version
kubectl version --client
```

devcontainer

## Notes
- Requires Docker and Docker Compose
- Initial container build may take 3-4 minutes


.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+      - NODE_ENV=development
+      # Path configuration
+      - PATH=/usr/local/go/bin:/go/bin:/usr/local/node/bin:${PATH}
+      # Build optimizations
+      - DOCKER_BUILDKIT=1
+
+    security_opt:
+      - seccomp=unconfined
+    cap_add:
+      - SYS_PTRACE
```
kyasbal, line 45: Could we know the purpose of them?

.devcontainer/setup-check.sh
```diff
@@ -0,0 +1,117 @@
...
+        echo "Install Docker Desktop to get Docker Compose"
+    elif [[ "$OSTYPE" == "linux-gnu"* ]]; then
+        echo "[INFO] Linux: Install Docker Compose plugin"
+        echo "1. Visit: https://docs.docker.com/compose/install/linux/"
+        echo "2. Follow the instructions to install Docker Compose plugin"
+    else
+        echo "[ERROR] Unsupported operating system"
+        exit 1
+    fi
+}
```
kyasbal, line 40: Could we consider using `docker compose` command instead of `docker-compose`?

Now the docker command contains `compose` subcommand natively.
https://docs.docker.com/compose/gettingstarted/#step-3-build-and-run-your-app-with-compose

Let me know if I didn't get your intention correctly.

.devcontainer/setup-dev.sh
```diff
@@ -0,0 +1,86 @@
...
+            rm -rf /tmp/cache
+            break
+        fi
+        log "Retry $i: npm install failed, retrying..."
+        sleep 5
+    done
+
+    # Install Angular CLI only if not already installed
+    if ! command -v ng &>/dev/null; then
+        npm install -g --prefix ~/.npm-global @angular/cli || {
```
kyasbal, line 37: Is `--prefix ~/.npm-global` needed here? I saw it's configured in the Dockerfile.

kyasbal (changes requested):
Thank you for contributing our code base! 

I haven't tried around `.devcontainer` for a while, but I learnt how this is well configured and make new contributors easy to start contributes. I think @RyuSA will also comment on this because he should be familiar with around this.

I added some comments. We are open to improve, thus please feel free to reply these suggestions if you have better solution rather than my suggestion.

Thanks! 👍

kyasbal:
And we need to add a license headers in the following scripts.

```
.devcontainer/docker-compose.yml
.devcontainer/DockerFile
.devcontainer/setup-dev.sh
.devcontainer/setup-check.sh
```

You can add this just by running `make add-licenses`
Ref: https://github.com/GoogleCloudPlatform/khi/blob/737a2f8e75d0a6a26207ed77dd884f9293baa066/scripts/make/codegen.mk#L21

.devcontainer/setup-check.sh
Sozhan308: Thank you for the feedback! You raise a good point about the [docker compose check](https://github.com/GoogleCloudPlatform/khi/pull/62/files#diff-058e34b818b62c84bc48247a6b2076e5dc34203f19ff96d83ce52b68c75101cbR90).

I agree we should standardize on `docker compose` (the integrated plugin approach) rather than checking for both versions. The Docker Compose plugin is now the officially recommended way to use Compose and comes pre-installed with Docker Desktop or `docker-compose-plugin` in linux https://docs.docker.com/compose/install/linux.

I've already updated our documentation to consistently use `docker compose` commands in the [README](https://github.com/GoogleCloudPlatform/khi/pull/62/files#diff-0f9548cbce10dbdef0b86d783ed0727a4ac1061fc71cb549b5dbd400705c71fdR18). The next step is to simplify our verification logic by removing the legacy docker-compose standalone check.

This aligns better with Docker's current best practices and simplifies our setup verification. I'll update the commit to reflect these changes.

.devcontainer/docker-compose.yml
Sozhan308: Regarding the security configurations in docker-compose.yml:

These development-only settings enable debugging capabilities:

1. `seccomp=unconfined`:
   - Disables the default seccomp (secure computing mode) profile
   - Allows unrestricted system calls needed by debuggers

2. `SYS_PTRACE` capability:
   - Enables process tracing capabilities (to inspect running processes)
   - Required for Go debugger (delve)

References:
- Docker Security Configuration: https://docs.docker.com/engine/reference/run/#security-configuration
- Docker Runtime Privileges: https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities
- Seccomp Security: https://docs.docker.com/engine/security/seccomp/

.devcontainer/DockerFile
```diff
@@ -0,0 +1,83 @@
...
+    libc6-dev \
+    linux-libc-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install development tools
+RUN set -ex \
+    # Install Go
+    && curl -fsSL "https://golang.org/dl/go${GO_VERSION}.linux-amd64.tar.gz" | tar -C /usr/local -xzf - \
+    # Install Node.js
+    && curl -fsSL "https://deb.nodesource.com/setup_${NODE_VERSION}.x" | bash - \
```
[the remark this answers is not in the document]
Sozhan308: Hey! Thanks for the feedback! 🙂

Based on your suggestion, I've added security checks to verify our downloads to make sure everything matches up before we install it. Would love to hear your thoughts on this approach! https://github.com/GoogleCloudPlatform/khi/pull/62/commits/90b5e6bb9437d7d1e48fc79323b2fb2558ff6df0

.devcontainer/setup-dev.sh
Sozhan308: Hey! Good catch! 👍

No need for the --prefix flag since we already set it up in the Dockerfile. I've removed it to keep things clean and avoid redundancy.

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,71 @@
...
+services:
+  dev-env:
+    build:
+      context: .
+      dockerfile: Dockerfile
+      args:
+        <<: *build-args
+        UBUNTU_VERSION: "22.04"
+        USER_UID: ${USER_UID:-1000}
+        USER_GID: ${USER_GID:-1000}
```
kyasbal, line 36: I wonder how these are common to have in environment variables.
I tested this with a new Debian environment just with a docker command. `$USER_UID` and `$USER_GID` don't exist but `$USERNAME` exists.

This seems to cause file permission error on `setup-dev.sh` L25 because these folder is hold by UID 1000, but VSCode seems to pass the right UID to connect the container.
It works if I set these outside of the container and we can document it, but do you have any better solution?

.devcontainer/docker-compose.yml
kyasbal: Noted! Thanks

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+    # Install kubectl with hash verification
+    && curl -LO "https://dl.k8s.io/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl" \
+    && echo "${KUBECTL_SHA256} kubectl" | sha256sum -c - \
+    && install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl \
+    && rm kubectl \
+    && rm -rf /var/lib/apt/lists/*
+
+# Create non-root user
+RUN groupadd --gid $USER_GID $USERNAME \
+    && useradd --uid $USER_UID --gid $USER_GID --shell /bin/bash --create-home $USERNAME \
```
kyasbal, line 94:
```suggestion
    && useradd -l --uid $USER_UID --gid $USER_GID --shell /bin/bash --create-home $USERNAME \
```

This is to avoid the same issue of this: https://github.com/moby/moby/issues/5419#issuecomment-41478290
I don't think it cause any side effect from my best knowledge.

My environment had very large UID...

kyasbal (changes requested):
I tested it on my newly setup Linux VM. It worked after configuring user id or group id environment variables as I pointed out in a review comment.

This devcontainer is very easy setup and so useful!

.devcontainer/DockerFile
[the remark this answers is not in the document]
kyasbal: Thank you for supporting hash verification! 
The method is not just only checking the hash from constant value, but also taking the value from the repo! It reduces our maintenance effort to keep these hash latest. It's super awesome.

.devcontainer/setup-check.sh
kyasbal: Thank you for fixing on this!

.devcontainer/docker-compose.yml
Sozhan308: Now, I've introduced a different approach https://github.com/GoogleCloudPlatform/khi/pull/62/commits/c1a6a47e40cb637cf61bca6b75d6a9411a49d2fd for the user handling to work better across different environments:

1. Made the GID allocation dynamic:
   - Uses next available GID incase of system group conflicts (like GID 20 for 'dialout')
   - Preserves specified GID when no conflicts exist
2. User setup is now more robust:
   - Used POSIX-standard `id` command for detection and UID/GID fallbacks in Dockerfile (UID/GID 1000)
   - Should ideally work with VSCode and other environments
 
I have tested and it's working. Let me know your thoughts on this!

kyasbal (changes requested):
Looks everything good except a thing.
The `.env` generated in `.devcontainer` after the build seems not to be ignored in `.gitignore`.

If I understand this change correctly, that should be ignored, shouldn't it?  I'm gonna approve this PR once this concern is solved.

Sozhan308:
> Looks everything good except a thing. The `.env` generated in `.devcontainer` after the build seems not to be ignored in `.gitignore`.
> 
> If I understand this change correctly, that should be ignored, shouldn't it? I'm gonna approve this PR once this concern is solved.

Yep, 👍  I'll add it

kyasbal (approved):
Looks awesome! Thank you a lot for your contribution 🎉 🎉 🎉

PR #63, opened 2025-02-08
Add reference generator builds reference from the task graphs 

kyasbal:
# Add reference generator builds reference from the task graphs

## Preview of the document: 

https://github.com/GoogleCloudPlatform/khi/blob/3621de61e8a8a132fd3c6f701f2f0907e3c41017/docs/en/reference/inspection-type.md

>[!IMPORTANT]
> Making the whole this reference to a perfect document takes time especially if we care about every sentences of English.
> Let's focus our reviews on the way of reference generation and the abstract structure of these reference regarding its template at this moment. And let's improve the content of them with another issue following this PR.

## Overview

This PR adds document generator that create/update document markdowns generated from KHI source code.
KHI backend is built top of a DAG task runner system and each tasks have labels. This change make use of these to generate these documents from source code.

This change aims to achieve the following outcomes.

* Reducing maintenance effort to keep codes and document latest
* Testability over document. We can write tests to check documents without description or any important parameters.
* Link from the frontend. We can generate anchor for headers automatically from values defined in the code base. It's easy to link these document from the corresponding UI element in future.

## Amendable document auto generator

This PR utilizes the Golang package `template/text` with considering the usage for contributors to extend documents.
Generated text from templates consists of multiple section surrounded by:

```
<!-- BEGIN GENERATED PART: <ID of the generated section> -->
...(actual contents)
<!-- END GENERATED PART: <ID of the generated section> -->
```

Users can add texts after `<!-- END GENERATED PART: <ID of the generated section> -->` and the document generator keep the position sticked to the previous generated section.

This approach will generate the document with a bit noisy comments, but it enables humans to add contents directly on the document. KHI documents is expected to  be hard if that was completely written only with texts, and several features have different points to pay attention. We needed the way to extend the automatically generated reference by human hand especially for adding screenshots, callouts or additional links ...etc.


kyasbal:
We are thinking about the other way, especially about thinking using [Hugo](https://gohugo.io/) to align with the doc on Kubernetes.

kyasbal:
@jyane Can we proceed review for this? At least for the document model part wound be better to have reviews, maybe?

jyane:
I'm extremely busy nowadays due to internal things, @RyuSA could you take a look?

RyuSA:
Let me merge this pull request. 👀

We still have a lot of features to add to KHI, and we need to improve the documentation to make it easier to search. We would like "documentation website" with a lightweight SSG (like Hugo) in the future, but we can't do that right away for various reasons.

This PR has everything we need, and it doesn't make any major changes to main.go, so we can easily revert it if we don't need the generated documentations anymore.

RyuSA (approved):
LGTM

PR #66, opened 2025-02-13
fix the issue graph page won't be rendered

kyasbal:
This seems to be happened after upgrading frontend dependencies.

Now it works!
diagram (3)


renamoo (approved):
LGTM

Issue #72, opened 2025-02-15
Support OSS kube-apiserver audit log


kyasbal:
## Summary

This change adds supporting the JSONline audit logs of OSS kube-apiserver. 
Users upload audit log text files on the `new inspection` dialogue and process them as KHI do in audit log parsers for GCP.

Steps:

* Support file upload on the `new inspection` dialog
* Add `OSS kubernetes` inspection type
* Add parser for OSS kubernetes

## Related:

#31 

Branch: `epic/oss-k8s-apiserver`

kyasbal:
I made some progress on this. I only implemented few features accomplished from audit logs compared to the parsers in GKE.
But now I successfully parsed JSONL kube-audit log exported from a KIND cluster. Both screenshots below are from the OSS kube-apiserver JSONL logs.

![Image](https://github.com/user-attachments/assets/9da08bc2-3ae9-41be-afd7-f1de423e0723)

![Image](https://github.com/user-attachments/assets/837ef213-9e80-4f31-9383-126df02ed4d7)

I still need to have more audit log especially not from KIND.

kyasbal:
We've launched this feature on the latest version today. I hope you try the feature on your cluster!
Please let us know if you encounter any issues in the feature. The environment should be varied by each clusters, we may need to consider some edge cases more. Your bug reports are valued!

![Image](https://github.com/user-attachments/assets/917a5fa9-5b47-4ff3-991d-37bcc817219a)

PR #73, opened 2025-02-15
Adding types to store uploaded files

kyasbal:
This pull request introduces a new type to manage uploaded files in an extensible way, laying the groundwork for supporting log input via file uploads in the future.

**Summary of Changes:**

We've created an `UploadFileStore` type to manage uploaded files, designed to be extensible for supporting file upload based inspection.

The upcoming file upload form element will utilize `UploadFileStore.GetUploadToken` to obtain a unique ID representing the uploaded file, based on specific parameters. This token will then be passed to the frontend.

The frontend will handle file uploads using the provided `UploadToken`.  This PR implements `LocalUploadFileStoreProvider`, which stores uploaded files in a specific directory on the server. The server should implement this upload mechanism in the later commit.  The design allows for future extensions, such as importing files uploaded to cloud storage services like GCS.

Once a file is uploaded, it's checked by the `UploadFileVerifier` specified during the upload process. This verifier confirms that the file meets the expected criteria.  For example, the included `JSONLineUploadFileVerifier` checks if the uploaded file is in valid JSON Lines format.

Throughout this upload and verification process, the `UploadFileStore` provides a way to retrieve the status associated with a token in a JSON-serializable format.  This enables these errors to be included in the form metadata to show them on frontend.

### DIFF
--- a/pkg/server/upload/store.go
+++ b/pkg/server/upload/store.go
@@ -57,15 +57,21 @@ type UploadFileStore struct {
 	results       map[string]UploadResult
 	verifierLock  sync.RWMutex
 	verifiers     map[string]UploadFileVerifier
+	tokenHashes   map[string]interface{}
+	tokenHashLock sync.RWMutex
 }
 
 // GetUploadToken returns the token to upload it from frontend.
+// The ID must be combination of a known string and random string to make it harder to guess it from outside.
 func (s *UploadFileStore) GetUploadToken(id string, verifier UploadFileVerifier) UploadToken {
 	s.resultLock.Lock()
 	s.verifierLock.Lock()
+	s.tokenHashLock.Lock()
 	defer s.resultLock.Unlock()
 	defer s.verifierLock.Unlock()
+	defer s.tokenHashLock.Unlock()
 	token := s.StoreProvider.GetUploadToken(id)
+	s.tokenHashes[token.GetHash()] = struct{}{}
 	_, ok := s.results[token.GetID()]
 	if !ok {
 		s.results[token.GetID()] = UploadResult{
@@ -78,9 +84,13 @@ func (s *UploadFileStore) GetUploadToken(id string, verifier UploadFileVerifier)
 
 // GetResult returns the result of the upload with given token.
 func (s *UploadFileStore) GetResult(token UploadToken) (UploadResult, error) {
+	err := s.ensureIssuedToken(token)
+	if err != nil {
+		return UploadResult{}, err
+	}
 	s.resultLock.RLock()
+	defer s.resultLock.RUnlock()
 	result, ok := s.results[token.GetID()]
-	s.resultLock.RUnlock()
 	if ok {
 		return result, nil
 	}
@@ -89,6 +99,10 @@ func (s *UploadFileStore) GetResult(token UploadToken) (UploadResult, error) {
 
 // SetResultOnStartingUpload sets the upload status to Uploading.  It returns an error if the token is not found.
 func (s *UploadFileStore) SetResultOnStartingUpload(token UploadToken) error {
+	err := s.ensureIssuedToken(token)
+	if err != nil {
+		return err
+	}
 	s.resultLock.Lock()
 	defer s.resultLock.Unlock()
 	_, ok := s.results[token.GetID()]
@@ -104,6 +118,10 @@ func (s *UploadFileStore) SetResultOnStartingUpload(token UploadToken) error {
 
 // SetResultOnCompletedUpload notify the file upload is completed and start verifier.
 func (s *UploadFileStore) SetResultOnCompletedUpload(token UploadToken, uploadError error) error {
+	err := s.ensureIssuedToken(token)
+	if err != nil {
+		return err
+	}
 	s.resultLock.Lock()
 	defer s.resultLock.Unlock()
 	prev, ok := s.results[token.GetID()]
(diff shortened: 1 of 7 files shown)


kyasbal:
This is a sub task of #72

pkg/server/upload/storeprovider.go
```diff
@@ -0,0 +1,93 @@
...
+// if it does not exist.  It returns an error if the directory
+// cannot be created (and does not already exist).
+func NewLocalUploadFileStoreProvider(directoryPath string) (*LocalUploadFileStoreProvider, error) {
+	// Create the directory (and any parent directories) if it doesn't exist.
+	// os.MkdirAll will not return an error if the directory already exists.
+	err := os.MkdirAll(directoryPath, 0755)
+	if err != nil {
+		return nil, err // Return the error if directory creation fails.
+	}
+	return &LocalUploadFileStoreProvider{directoryPath: directoryPath}, nil
```
jyane, line 54: I think a constructor has side-effects is not really great in general. Can we move this to Write?

pkg/server/upload/storeprovider.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// NewLocalUploadFileStoreProvider creates a new LocalUploadFileStore.
+// It takes the directory path as an argument and creates the directory
+// if it does not exist.  It returns an error if the directory
+// cannot be created (and does not already exist).
+func NewLocalUploadFileStoreProvider(directoryPath string) (*LocalUploadFileStoreProvider, error) {
+	// Create the directory (and any parent directories) if it doesn't exist.
+	// os.MkdirAll will not return an error if the directory already exists.
+	err := os.MkdirAll(directoryPath, 0755)
```
jyane, line 50: Any specific reason to provide read/exec permissions?

pkg/server/upload/storeprovider.go
```diff
@@ -0,0 +1,93 @@
...
+	return &LocalUploadFileStoreProvider{directoryPath: directoryPath}, nil
+}
+
+// GetUploadToken implements UploadFileStoreProvider.
+func (l *LocalUploadFileStoreProvider) GetUploadToken(id string) UploadToken {
+	return &DirectUploadToken{ID: id}
+}
+
+func (l *LocalUploadFileStoreProvider) Read(token UploadToken) (io.ReadCloser, error) {
+	filePath := filepath.Join(l.directoryPath, token.GetID())
```
jyane, line 63: The token validity would be checked in somewhere before here?
Even so, this may be a bit dangerous if end users can send a request with any arbitrary token.

pkg/server/upload/storeprovider.go
```diff
@@ -0,0 +1,93 @@
...
+		if os.IsNotExist(err) {
+			return nil, os.ErrNotExist
+		}
+		return nil, err
+	}
+	return file, nil // os.File implements io.ReadCloser
+}
+
+func (l *LocalUploadFileStoreProvider) Write(token UploadToken, reader io.Reader) error {
+	filePath := filepath.Join(l.directoryPath, token.GetID())
```
jyane, line 75: Ditto.

pkg/server/upload/verifier.go
```diff
@@ -0,0 +1,76 @@
...
+		lineNumber++
+		line := scanner.Bytes()
+
+		// Check if the line is empty or just whitespace. Skip empty lines.
+		if len(bytes.TrimSpace(line)) == 0 {
+			continue
+		}
+
+		var data interface{}
+		if err := json.Unmarshal(line, &data); err != nil {
```
jyane, line 64: Here is a bit weird to check the JSON validity?

https://pkg.go.dev/encoding/json#Valid

pkg/server/upload/store.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	s.verifiers[token.GetID()] = verifier
+	return token
+}
+
+// GetResult returns the result of the upload with given token.
+func (s *UploadFileStore) GetResult(token UploadToken) (UploadResult, error) {
+	s.resultLock.RLock()
+	result, ok := s.results[token.GetID()]
+	s.resultLock.RUnlock()
```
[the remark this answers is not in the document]
kyasbal: Done

pkg/server/upload/storeprovider.go
kyasbal: I changed it to `0700` because we don't need to grant the other to access the uploaded file.
It's a permission for directory exec permission is needed for writing the file and it provides read functionality using the read permission.

pkg/server/upload/storeprovider.go
kyasbal: Done

pkg/server/upload/storeprovider.go
kyasbal: Done.

pkg/server/upload/storeprovider.go
kyasbal: Done.

pkg/server/upload/verifier.go
kyasbal: I changed the logic to use json.Valid instead, but I needed to deserialize it to get the error message.

jyane: Oh, true. So it's up to you. I don't think we need Valid here... I'm sorry I didn't think about the error message.

jyane (approved):
LGTM

Issue #57, opened 2025-02-04
Add a flag to skip tests using Cloud Logging API


kyasbal:
We need few weeks to get Workload Identity Pool in our testing project and we can't enable backend tests using Cloud Logging on CI until that.

Let's make a flag to skip these tests for running these tests in current CI without using Cloud Logging.

Okabe-Junya:
@kyasbal 

Hello, and first of all, thank you for releasing such amazing software!

I'm very interested in this software, and when I tried to hack it, I noticed that running `make test` (i.e. `go test ./...`) fails. So, I'd like to work on this issue.

There are several ways to operate on the files under test in Go, but which method would be best (or another)?

1. Create an `internal/testflags` package and blank-import it in every package that runs tests (treating the flags as global variables; otherwise, it probably won’t be possible to run it as `go test ./... -args <myflag>=true`).
2. Use the built-in short (`go test -short`) flag.
3. Use environment variables as flags.

If this issue is only a temporary change until the Workload Identity Pool becomes available, then using environment variables would be the simplest approach. 
However, if you plan to continue running `make test` (or `make test-local`) in the development environment going forward, creating a dedicated package for test flags would be worthwhile.

Thanks!

kyasbal:
@Okabe-Junya Thank you for being interested in contributing this project 🙇 

> There are several ways to operate on the files under test in Go, but which method would be best (or another)?
> Create an internal/testflags package and blank-import it in every package that runs tests (treating the flags as global variables; otherwise, it probably won’t be possible to run it as go test ./... -args <myflag>=true).
> Use the built-in short (go test -short) flag.
> Use environment variables as flags.

> If this issue is only a temporary change until the Workload Identity Pool becomes available, then using environment variables would be the simplest approach.
> However, if you plan to continue running make test (or make test-local) in the development environment going forward, creating a dedicated package for test flags would be worthwhile.

I prefer the first way. It's because we'll support OSS kubernetes logs or other log storage in future and it needs to call some APIs not on Google Cloud. In the case, I also want to disable calling it in my test because I don't want to prepare every environment.
This would be same for contributor only interested in OSS kubernetes log feature part and wanting to ignore tests strictly related to Google Cloud. It's should be okey eventually tested in CI before merging. We should keep capability to disable them in local environment.

[...]

Issue #29, opened 2025-01-30
Setup Github Action to test backend


kyasbal:
Proceed the configuration after getting the Workload Identity Pool for this project.

Okabe-Junya:
(This may be temporary, but) I've enabled the unit tests that can run without Cloud Logging in #78, thanks!!

kyasbal:
@kyasbal TODO: Create another workload identity pool configuration issue and close this issue.

kyasbal:
Close this for now. The later remaining tasks will be continued after we getting the pool. #111

PR #78, opened 2025-02-20
Chore: Enable unit test ci with skipping flag

Okabe-Junya:
ref. #29, #57 

## What

- Enable unit test CI _**without using**_ cloud logging

## Background

In #57, a flag was implemented to enable running only unit test that does not use cloud logging. 
With this flag, it has become possible to run unit test CI on PR (`on.pull_request`).

### DIFF
--- a/.github/workflows/pr-request.yaml
+++ b/.github/workflows/pr-request.yaml
@@ -64,8 +64,7 @@ jobs:
       run: |
         mkdir -p dist/session
         # A placeholder frontend code read from backend test
-        echo "" > ./dist/index.html
-        echo "" > ./dist/session/100
+        echo '<!--INJECT GENERATED CODE HERE FROM BACKEND-->' > ./dist/index.html
         go test ./... -args -skip-cloud-logging=true
 
     - name: Backend Format and Lint Check
--- a/.github/workflows/pr-request.yaml
+++ b/.github/workflows/pr-request.yaml
@@ -58,6 +58,11 @@ jobs:
     #     echo "" > ./dist/index.html
     #     make coverage-go
 
+    # For now, we skip the cloud logging test
+    # TODO: Remove the flag after the cloud logging test is ready
+    - name: Backend Test
+      run: go test ./... -args -skip-cloud-logging=true
+
     - name: Backend Format and Lint Check
       run: |
         make check-format-go
--- a/.github/workflows/pr-request.yaml
+++ b/.github/workflows/pr-request.yaml
@@ -62,10 +62,10 @@ jobs:
     # TODO: Remove the flag after the cloud logging test is ready
     - name: Backend Test
       run: |
-        mkdir dist
+        mkdir -p dist/session
         # A placeholder frontend code read from backend test
         echo "" > ./dist/index.html
-        make build-web
+        echo "" > ./dist/session/100
         go test ./... -args -skip-cloud-logging=true
 
     - name: Backend Format and Lint Check
--- a/.github/workflows/pr-request.yaml
+++ b/.github/workflows/pr-request.yaml
@@ -65,6 +65,7 @@ jobs:
         mkdir dist
         # A placeholder frontend code read from backend test
         echo "" > ./dist/index.html
+        make build-web
         go test ./... -args -skip-cloud-logging=true
 
     - name: Backend Format and Lint Check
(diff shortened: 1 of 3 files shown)


Okabe-Junya:
Do we need `make build-web` before go test?

kyasbal:
> Do we need make build-web before go test?

I feel it's a bit better to create a empty fake file in the directory.
Because the test failure on the Github Action task should mean the test failure on the backend only as much as possible, if we add `make build-web` step in there, the task will fail with frontend problem.

How do you think?

kyasbal:
> Looking at the CI error log, it appears that we need two files: `dist/index.html` and `dist/session/100`.

hmm.. that's weird. The viewer mode should return the index.html for session/*

.github/workflows/pr-request.yaml
```diff
@@ -58,6 +58,16 @@ jobs:
...
     #     make coverage-go
 
+    # For now, we skip the cloud logging test
+    # TODO: Remove the flag after the cloud logging test is ready
+    - name: Backend Test
+      run: |
+        mkdir -p dist/session
+        # A placeholder frontend code read from backend test
+        echo "" > ./dist/index.html
+        echo "" > ./dist/session/100
```
kyasbal, line 68:
```suggestion
```

I think this is not needed. I tried it on my environment without `./dist/session/100` and the tests passed all. 
Can we find why these happens?

More precisely, request against `sessions/100` shouldn't use the static file `./dist/session/100`, but it should just return the index.html instead.
https://github.com/GoogleCloudPlatform/khi/blob/main/pkg/server/server.go#L72

kyasbal (changes requested):

.github/workflows/pr-request.yaml
Okabe-Junya: Thanks!!

I thoroughly investigated this test error. In my local environment and in the GitHub Actions for the forked repo, setting the placeholder in `dist/index.html` wasn’t enough to pass the test.

ref. https://github.com/Okabe-Junya/khi/actions/runs/13484508992/job/37674161252

server.go shows that [`static.Serve`](https://github.com/gin-contrib/static/blob/master/serve.go) engine is being used.

https://github.com/GoogleCloudPlatform/khi/blob/7e35fb69714e2d92f7a6c916df1b81d3d28a4eb8/pkg/server/server.go#L66

That is, if a file matching the request path (in this case, `/session/100`) exists, it is returned this file, and the status code is 200.  
(This is why the CI for 71b3b8f passed)

On the other hand, if no file matching the request path exists, it appears that the `replaceDynamicPartOfIndex` function is called.

https://github.com/GoogleCloudPlatform/khi/blob/7e35fb69714e2d92f7a6c916df1b81d3d28a4eb8/pkg/server/server.go#L80

https://github.com/GoogleCloudPlatform/khi/blob/7e35fb69714e2d92f7a6c916df1b81d3d28a4eb8/pkg/server/index.go#L34-L37

In this function, it seems that an error is returned if the HTML does not include the argument `IndexReplacePlaceholder`. 
Since an empty file created with `echo ""` does not contain the IndexReplacePlaceholder, a 500 error occurs — I think the unit test will fail.

If my hypothesis is correct, it may be necessary to write some basic HTML in index.html. What do you think?

kyasbal: You are totally right. I forgot to have written `echo "<!--INJECT GENERATED CODE HERE FROM BACKEND-->" > ./dist/index.html` in the CI configuration to test backend codes before OSSing this.

Okabe-Junya: Thank you!!

It appears that we're ready to move forward with this PR!!
I fixed index.html in 0586a8f, and the CI (go test) passed.

ref. https://github.com/Okabe-Junya/khi/actions/runs/13528732497/job/37805609182

It passed go test on my laptop as well.

Please take another look, thanks!

kyasbal:
Changes are looks great. But there were several test packages added after creating this PR.
Thus the CI is failing. Could you do `git rebase origin/main` to rebase this changes on the top of main chain?

And you would need to add 2 blank imports

* pkg/document/generator/generator_test.go
* pkg/inspection/task/label/query_test.go

Okabe-Junya:
thanks!!

I updated some tests in bd62fea, please take another look!!

kyasbal (approved):
Thanks! Now the all tests are passing. Thank you a lot for your contribution. This made our code base much more stable and maintainable 🥇

PR #81, opened 2025-02-23
Added a new types to define form fields for parameters

kyasbal:
This PR creates several new types in `web/app/src/common/schema/form-types.ts`. KHI generates form fields from the JSON data given from backend, the form field used to only support text input but this change expand this limitation to support multiple types of input forms.

**Done in this PR**

* Defined the types of form fields settings in `web/app/sr/common/schema/form-types.ts`.
* Renamed `file-upload.component.*` to `file-parameter.component.*`
* Created common components used in parameter forms (hint & header)
* Defined `text-parameter.component.*


web/src/app/common/schema/form-types.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Text type parameter specific data in ParameterFormField type.
+ */
+export interface TextParameterFormField extends ParameterFormFieldBase {
+  type: ParameterInputType.Text;
+  /**
+   * If this text form field is editable from user or not.
+   */
+  allowEdit: boolean;
```
renamoo, line 73: Not a strong opinion so it's totally fine to keep this name. With that being said, just sharing my thoughts that the popular naming for this type of property might be "readonly"
https://developer.mozilla.org/ja/docs/Web/HTML/Attributes/readonly

web/src/app/common/schema/form-types.ts
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Label of the parameter.
+   */
+  label: string;
+  /**
+   * Description of this parameter.
+   */
+  description: string;
+  /**
+   * Type of hint message (error,warning or info)
```
renamoo, line 56: Super nit: `(error,warning or info)` -> `(error, warning or info)`

renamoo (approved):
Overall LGTM!

PR #84, opened 2025-02-26
Replace new parameter form system

kyasbal:
We have been worked on improving parameter form architecture in the frontend. This includes the changes around schema used in response of the `dryrun` endpoint which performs mostly validation logics on the form.

By this change, I replaced the long DOMs used in `new-inspection` dialog with the components having been working on, and migrate the backend form builders to use the newer schema.


kyasbal:
The newer parameter form.

image

kyasbal:
This PR mixes the changes on frontend and backend because it's for migrating the schema.
I'm sorry to ask you this relatively large PR but hopefully you could review the part where you can.

web/src/app/dialogs/new-inspection/components/service/parameter-store.ts
```diff
@@ -88,8 +93,11 @@ export class DefaultParameterStore implements ParameterStore {
         take(1),
       )
       .subscribe((parameters) => {
-        parameters[id] = value;
-        this.currentParameters.next(parameters);
+        this.currentParameters.next({
+          ...parameters,
+          [id]: value,
+        });
+        console.log('parameter update', value);
```
renamoo, line 100: Is it intended to keep this log?

web/src/app/dialogs/new-inspection/components/service/parameter-store.ts
```diff
@@ -107,4 +115,24 @@ export class DefaultParameterStore implements ParameterStore {
   public destroy(): void {
     this.destroyed.next(void 0);
   }
+
+  private deepEqual(
```
renamoo, line 119: It's technically not deep equal because it does not compare value deeply. However, it should work as long it's assured that values in object never be object or array etc. We might want to add comment about this limitation if so.

web/src/app/dialogs/new-inspection/new-inspection.component.ts
```diff
@@ -249,26 +239,92 @@ export class NewInspectionDialogComponent implements OnDestroy {
...
+          break;
+      }
+    }
+    return result;
   }
 
-  public run() {
-    this.runRequest.next(null);
+  /**
+   * Count the count of fields with error.
```
renamoo, line 293: nit: maybe simply say "Count error fields" or "Count fields that contain errors" or something like that?

web/src/app/dialogs/new-inspection/new-inspection.component.ts
```diff
@@ -249,26 +239,92 @@ export class NewInspectionDialogComponent implements OnDestroy {
...
+        result += this.countErrorFields(parameter.children);
+      } else if (parameter.hintType === ParameterHintType.Error) {
+        result++;
+      }
+    }
+    return result;
+  }
+
+  /**
+   * Count the count of all fields.
```
renamoo, line 309: nit: Same as above

web/src/app/dialogs/new-inspection/new-inspection.component.ts
```diff
@@ -99,15 +116,24 @@ type FormFieldViewModel = {
...
   ],
 })
 export class NewInspectionDialogComponent implements OnDestroy {
-  private destoroyed = new Subject<void>();
-
   static readonly STEP_INDEX_CLUSTER_TYPE = 0;
   static readonly STEP_INDEX_FEATURE_SELECTION = 1;
   static readonly STEP_INDEX_PARAMETER_INPUT = 2;
 
+  private destoroyed = new Subject<void>();
```
renamoo, line 133: nit: destroyed

web/src/app/dialogs/new-inspection/new-inspection.component.ts
```diff
@@ -249,26 +239,92 @@ export class NewInspectionDialogComponent implements OnDestroy {
...
     }
   }
 
   public toggleFeature(featureId: string) {
     this.featureToggleRequest.next(featureId);
   }
 
+  public run() {
+    this.hadRun.set(true);
+    of(void 0)
```
renamoo, line 252: I guess `of(void 0)` is used to trigger this flow only when the "Run" button is clicked but it might not be intuitive. 

Potential alternative:

```
(click)="run()"

private startInspectionSubject = new Subject<void>();

run() {
    this.startInspectionSubject.next();
}

ngOnInit() {
    this.startInspectionSubject.pipe(
        takeUntil(this.destroyed),
        withLatestFrom(
....
```

or `combineLatestWith` but there is a slight difference in a behavior so it depends on the expected timeline of observable emission

web/src/app/dialogs/new-inspection/components/service/parameter-store.ts
kyasbal: Thanks. Removed.
I'm thinking if there were good way to check this with linter.

web/src/app/dialogs/new-inspection/components/service/parameter-store.ts
kyasbal: Renamed it to `haveEqualKeyValues`

web/src/app/dialogs/new-inspection/new-inspection.component.ts
kyasbal: Fixed

web/src/app/dialogs/new-inspection/new-inspection.component.ts
kyasbal: Fxied

web/src/app/dialogs/new-inspection/new-inspection.component.ts
kyasbal: Fixed

web/src/app/dialogs/new-inspection/new-inspection.component.ts
kyasbal: Nice. I've followed your suggestion.

web/src/app/dialogs/new-inspection/new-inspection.component.ts
```diff
@@ -175,7 +189,7 @@ export class NewInspectionDialogComponent implements OnDestroy {
   public currentTaskClient = this.currentInspectionType.pipe(
     filter((type) => !!type),
     switchMap((taskType) => this.apiClient.createInspection(taskType!.id)),
-    share(),
+    shareReplay(1),
```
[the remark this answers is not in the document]
kyasbal: I inserted `takeUntil(this.destroyed)` in the pipe.

`refCount` can be one of the solution, but I believe it's better when we know the timing of the observable becoming no longer needed.

pkg/inspection/metadata/form/form.go
```diff
@@ -20,36 +20,79 @@ import (
...
-	Hint            string            `json:"hint"`
-	HintType        FormFieldHintType `json:"hintType"`
-	Default         string            `json:"default"`
-	AllowEdit       bool              `json:"allowEdit"`
-	Suggestions     []string          `json:"suggestions"`
-	ValidationError string            `json:"validationError"`
+// ParameterHintType represents the types of hint message shown at the bottom of parameter forms.
+type ParameterHintType string
+
+const (
```
RyuSA, line 41: Could you add comments to this Enum? Without comments, it's unclear how each value is used in the frontend. `None` and `Info` are very similar.

- None:  The user can use the form as is.  Use this value when there are no particular hints.
- Info:  The user can use the form as is, but there are helpful hints for input (e.g., the calculated time range or query).
- Warn: The user can use the form as is, but there are points to be aware of when entering input (e.g., the entered cluster not found).
- Error: The user cannot use the form as is.  Prompt the user to make corrections based on the value of the hint.

RyuSA (commented):
almost LGTM(allowEdit -> readOnly is a little bit tough...)
I left a comment for `comment`.

web/src/app/dialogs/new-inspection/components/service/parameter-store.ts
renamoo: Maybe we can use ESLint "no-console" rule and `ng lint` ?
https://eslint.org/docs/latest/rules/no-console
https://angular.jp/cli/lint

renamoo (approved):
LGTM from myside 👋

pkg/inspection/metadata/form/form.go
kyasbal: fixed

RyuSA (approved):
LGTM

PR #87, opened 2025-03-05
Add Angular proxy configuration to simplify development workflow

kyasbal:
## Overview
This PR adds proxy configuration to the Angular development server, integrating the frontend and backend development environments. This eliminates the need to manage two different ports (Angular: 4200, Golang: 8080) separately during development.

## Changes
* Added `web/proxy.conf.mjs` file implementing proxy configuration that forwards requests to the `/api` path to `localhost:8080`
* Updated angular-template.json file to reference the proxy configuration
* Additionally fixed the retry settings in the logic to get the configruation from backend. (I've found massive retry request made server and frontend extremely heavy when I run frontend without starting the backend)

## References

https://angular.dev/tools/cli/serve#proxying-to-a-backend-server

### DIFF
--- a/docs/en/development-guide.md
+++ b/docs/en/development-guide.md
@@ -68,7 +68,7 @@ To develop frontend code, we usually start Angular dev server on port 4200 with 
 $ make watch-web
 ```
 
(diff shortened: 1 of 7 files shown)


renamoo (approved):
Looks very useful!

PR #94, opened 2025-03-12
Perf/require inspection types on features

kyasbal:
# Background

The log parse system in KHI is a DAG based system. KHI uses labels attached on tasks to organize and filter them.
On the inspecting a new cluster, users will select inspection type at first and KHI filters tasks supported by the inspection type using the label. If a task had no inspection type label, KHI assumed the task is supported by any of inspection types.

However, now we are expanding support to OSS clusters. There is no common format used in any cluster types regarding expanding KHI to OSS. Now every feature tasks are required to have inspection types.

# Changes

* The `FeatureTaskLabel` now receives array of IDs of supported inspection types.
* Defined several constant arrays of inspection type IDs in `pkg/source/gcp/inspectiontype/inspection_type.go`.


RyuSA (approved):
LGMT

PR #112, opened 2025-03-28
Adding document about KHI task system

kyasbal:
Add the task system document. This would be the fundamental part of the development guide for KHI.


kyasbal:
I will fix some part of this documents to align with current implementation in this weekend.
It's not sufficient but I'll merge it after the fix.

RyuSA (approved):
LGTM

Issue #107, opened 2025-03-28
Replace the feature selection API with PATCH from the PUT API


kyasbal:
# Background

Feature selection screen requests the server with the whole map containing enable flag of each log types.
![Image](https://github.com/user-attachments/assets/2a0a291f-ef6a-48e1-af3a-3a50d2d03123)

This can be flappy when user selected/deselected the query feature faster than sending its status to its backend.

# Approach

Replace `PUT /api/v2/inspection/tasks/:taskId/features` API with `PATCH /api/v2/inspection/tasks/:taskId/features`.

Current schema of the put request

```go
type PutInspectionTaskFeatureRequest struct {
	Features []string `json:"features"`
}
```

This should be replaced with the following schema.

```go
type PatchInspecionTaskFeatureRequest struct {
	Features map[string]bool `json:"features"`
}
```

azgartaj:
Hey @kyasbal , I'd like to take up this bug. Can you assign it to me?

kyasbal:
@azgartaj Yes! Thank you so much!

kyasbal:
I'll write a little more context just for assisting you.

This is the currently used PUT API. This should be rewritten with the PATCH API.
https://github.com/GoogleCloudPlatform/khi/blob/main/pkg/server/server.go#L144-L162

Currently, the implementation calls SetFeatureList on the runner with the provided name list of features.
But this would be rewritten with using the map in PatchInspecionTaskFeatureRequest and FeatureList method to get the current  enabled features.
https://github.com/GoogleCloudPlatform/khi/blob/main/pkg/inspection/runner.go#L133

We also need to rewrite some codes in API clients.
https://github.com/GoogleCloudPlatform/khi/tree/main/web/src/app/services/api

And it's called from only here.
https://github.com/GoogleCloudPlatform/khi/blob/main/web/src/app/dialogs/new-inspection/new-inspection.component.ts#L118-L133

---

I'm sorry that there are some of my poor quality codes around the part. If you are not familiar with frontend part, you can just additionally implement the PATCH API. I can do for the frontend part.

azgartaj:
Thank you for providing additional context! 
I have a little experience with frontend ,so I'll try figure that out as well. 
Is any release close-by that we are targeting this change for?

kyasbal:
This is a bug fix only for the environment having low speed connection against KHI server. This is impacting UX in these environment, but it's not critical at this moment.

[...]

PR #114, opened 2025-03-30
Rename types around DAG task system

kyasbal:
**This PR is not changing anything in logic. **

This changes fundamental type names to address these functionality precisely.

* `task.UntypedDefinition` -> `task.UntypedTask`
* `task.Definition[T]` -> `task.Task[T]`
* `task.DefinitionSet` -> `task.TaskSet`
* `task_test.MockTask` -> `task_test.StubTask`
* `task_test.MockTaskFromReferenceID` -> `task_test.StubTaskFromReferenceID`

PR #116, opened 2025-03-30
Enhancement: features page to use PATCH instead of PUT to update

azgartaj:
image

Features page now uses PATCH with only the updated feature id to update the state of selected features. The UI has been loaded and tested with changes included.
Issue: #107


pkg/inspection/runner.go
```diff
@@ -154,6 +154,25 @@ func (i *InspectionRunner) SetFeatureList(featureList []string) error {
 	return nil
 }
 
+func (i *InspectionRunner) UpdateFeatureMap(featureMap map[string]bool) error {
```
kyasbal, line 157: Can you add a comment for the method please?
(I'm sorry to ask this even I couldn't follow the rule for the most of the other existing code. I would like to add comments for the all public types, methods later.)

pkg/task/definition_set.go
```diff
@@ -75,6 +75,27 @@ func (s *DefinitionSet) Add(newTask UntypedDefinition) error {
 	return nil
 }
 
+// Remove a task definition from current DefinitionSet.
+// Returns error if the definition does not exist
+func (s *DefinitionSet) Remove(id string) error {
```
kyasbal, line 80: Could you add a simple unit test for this `Remove` method in `definition_set_test.go`?

web/src/app/services/api/backend-api.service.ts
```diff
@@ -147,12 +147,12 @@ export class BackendAPIImpl implements BackendAPI {
     return this.http.get<GetInspectionTaskFeatureResponse>(url);
   }
 
-  public setEnabledFeatures(taskId: string, featureIds: string[]) {
+  public setEnabledFeatures(taskId: string, featureIds: {[key: string]: boolean}) {
```
kyasbal, line 150: I wonder if the argument name `featureIds` is not suitable for now. Could we change the name of argument for something suitable?

web/src/app/dialogs/new-inspection/new-inspection.component.ts
```diff
@@ -184,10 +180,13 @@ export class NewInspectionDialogComponent implements OnDestroy {
 
   public currentEnabledFeatures = this.currentTaskFeatures.pipe(
```
kyasbal, line 181: Are there any better name rather than `currentEnabledFeatures`?
This observable emits map of task id and it's enablement status. It can contain disabled task IDs as well.

kyasbal (changes requested):
Thank you for your blazingly fast contribution.

I commented some part of them and I'm sorry to ask you to do something that I couldn't do for all of the other codes.

pkg/inspection/runner.go
azgartaj: No problem at all. Happy to add :)
Done.

pkg/task/definition_set.go
azgartaj: Missed to notice there was a test case for this. 
Added.

web/src/app/services/api/backend-api.service.ts
azgartaj: Agree. Updated to `featureMap` instead of `featureIds`.

web/src/app/dialogs/new-inspection/new-inspection.component.ts
azgartaj: Updated the name to `featureStatusMap` , let me know if that sounds okay :)
but again, something I relate to, 
> naming is hardest part of programming

kyasbal (changes requested):
Thank you for your quick fix on them.
My commented points are fixed well thoughtfully.

These 2 CI seems to fail because of lint error.
* frontend-tests(https://github.com/GoogleCloudPlatform/khi/actions/runs/14165581483/job/39703306405?pr=116)
* backend-tests(https://github.com/GoogleCloudPlatform/khi/actions/runs/14165581483/job/39703307557?pr=116)

You don't need to manually fix them usually, but please consider run the command to setup git hook.
https://github.com/GoogleCloudPlatform/khi/blob/main/docs/en/development-guide.md#setup-git-hook

The hook automatically run formatter just before your commit. 

You can ignore the failing markdown CI part. These will be fixed in another PR and you didn't change anything changing markdown.

---
Here is another thing. We only accepts commit with a valid signature.
image

Please configure your git client to use a registered pub key to sign your commits.
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

kyasbal:
I'm sorry to cause conflicts but we renamed some non intuitive names in our base types.
https://github.com/GoogleCloudPlatform/khi/pull/114

Now `task.Definition` was renamed to `task.Task`. I think this change introduce some conflicts on this PR. 
I'm sorry for bothering you but I wish if you could merge this regarding this change.

azgartaj:
That explains the merge conflict I got :D . No problem at all @kyasbal , I'll make changes accordingly.

kyasbal:
We need to fix the following errors.

```
Error: pkg/task/definition_set_test.go:476:22: undefined: UntypedDefinition
Error: pkg/task/definition_set_test.go:477:3: undefined: newDebugDefinition
Error: pkg/task/definition_set_test.go:478:3: undefined: newDebugDefinition
```

`UntypedDefinition` was renamed to `UntypedTask`. `newDebugDefinition` was renamed to `newDebugTask`.

---

And thank you for setting up your git client to sign commits.
image

But the public key seems not to be registered on Github yet. 
https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key

kyasbal (approved):
LGTM. Thank you @azgartaj for your contribution 👍

RyuSA (approved):
LGTM

PR #118, opened 2025-04-01
Support multi-platform build

### LINKED ISSUE #105: Build container image as multi arch container image
Currently KHI only provides amd64 container image. We will support arm64 arch images.

RyuSA:
Close #105


kyasbal (approved):

RyuSA:
Note: I think the workflow CAN work...but if there are some issues I'll fix them.

PR #128, opened 2025-04-07
Update VERSION

kyasbal:
We forgot to update the VERSION after merging the commit to support OSS kube-apiserver log.


RyuSA (approved):
LGTM

PR #129, opened 2025-04-07
k8s audit task graph bug fix on GKE

kyasbal:
The GKE k8s audit log parser in the main branch was broken because the GKE k8s audit log parser task wrongly depends on GKE audit log. Fixed the dependency relationships on the task definitions.


RyuSA (approved):
LGTM

PR #133, opened 2025-04-10
Separated Apache Airflow Scheduler from the composer package

RyuSA:
In the current package structure, the boundary between Composer and Airflow is ambiguous. I plan to add status information for Composer itself in the future (such as database errors and system logs), so I'm separating them now.

**gcp/tasks/.../composer/**
* Registers all Cloud Composer events. Airflow registers Tasks as "one of Composer's events".
* Depends on Airflow tasks. 

**apache-airflow/**
* Processes logs for Airflow itself. **Does not depend on Composer.**
* Provides only parsers; does not provide log query task and register with Inspection Type.

This PR is a "template" before all migrations.


pkg/model/history/resourcepath/apacheairflow.go
```diff
@@ -18,20 +18,20 @@ import (
...
-// composer#taskinstance#DAGID#RUNID#TASKID-MAPINDEX
-func ComposerTaskInstance(ti *model.AirflowTaskInstance) ResourcePath {
+// airflow#taskinstance#DAGID#RUNID#TASKID-MAPINDEX
+func AirflowTaskInstance(ti *model.AirflowTaskInstance) ResourcePath {
 	var detail = ti.TaskId()
 	if ti.MapIndex() != "-1" {
 		detail += "+" + ti.MapIndex()
 	}
-	return SubresourceLayerGeneralItem("Cloud Composer", "Task Instance", ti.DagId(), ti.RunId(), detail)
+	return SubresourceLayerGeneralItem("Apache Airflow", "TaskInstance", ti.DagId(), ti.RunId(), detail)
```
kyasbal, line 27: If the resource is a subresource, you may need to consider adding `ParentRelationship` of it.
https://github.com/GoogleCloudPlatform/khi/blob/main/pkg/model/history/resourcepath/controlplane.go#L27

kyasbal (approved):
Approved. But I added a comment regarding the subresource resource path.

It should work with the current code, but it would be better to have the relationship label in subresources because the layer can be associated to its parent with various reasons.
The label helps user to understand why the resource is under the parent.

kyasbal (approved):

RyuSA:
image
Ok, I added a label for each task. it looks better :)

PR #139, opened 2025-04-15
Update composer ux

RyuSA:
Upgrade UX for Cloud Composer taks

- deprecated v1, added v3
- field order(following the data flow)
- implements autocompletes(location, gke cluster for composer)

### DIFF
--- a/pkg/source/gcp/api/gcp_client.go
+++ b/pkg/source/gcp/api/gcp_client.go
@@ -158,16 +158,30 @@ func (c *GCPClientImpl) CreateGCPHttpRequest(ctx context.Context, method string,
  * Get the list of GKE cluster names.
  */
 func (c *GCPClientImpl) GetClusterNames(ctx context.Context, projectId string) ([]string, error) {
-	type gkeCluster struct {
-		Name string `json:"name"`
+
+	clusters, err := c.GetClusters(ctx, projectId)
+	if err != nil {
+		return nil, err
 	}
-	type clusterListResponse struct {
-		Clusters      []*gkeCluster `json:"clusters"`
-		NextPageToken string        `json:"nextPageToken"`
-	}
+
 	var result []string
+	for _, cluster := range clusters {
+		result = append(result, cluster.Name)
+	}
+	return result, nil
+}
+
+// Get all GKE clusters from container.googleapis.com
+// ref: https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters/list
+func (c *GCPClientImpl) GetClusters(ctx context.Context, projectId string) ([]Cluster, error) {
+
+	type clusterListResponse struct {
+		Clusters      []*Cluster `json:"clusters"`
+		NextPageToken string     `json:"nextPageToken"`
+	}
 	pc := NewPageClient[clusterListResponse](c.BaseClient)
 	clusterListResponses, err := pc.GetAll(ctx, func(hasToken bool, nextPageToken string) (*http.Request, error) {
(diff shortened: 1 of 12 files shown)


pkg/source/gcp/task/cloud-composer/form/autocomplete.go
```diff
@@ -29,44 +27,69 @@ import (
...
+
 	client, err := api.DefaultGCPClientFactory.NewClient()
 	if err != nil {
 		return inspection_cached_task.PreviousTaskResult[*gcp_task.AutocompleteClusterNameList]{}, err
 	}
 
 	projectID := task.GetTaskResult(ctx, gcp_task.InputProjectIdTaskID.GetTaskReference())
-	if projectID != "" && projectID == prevValue.DependencyDigest {
+	environment := task.GetTaskResult(ctx, composer_taskid.InputComposerEnvironmentTaskID.GetTaskReference())
+	dependencyDigest := environment
```
kyasbal, line 44: The `dependencyDigest` should be concatenate of `environment` and `projectID`.
Because the cache must be refreshed when users modified project ID.

kyasbal (changes requested):

pkg/source/gcp/api/gcp_client.go
```diff
@@ -539,6 +554,45 @@ func (c *GCPClientImpl) GetComposerEnvironmentNames(ctx context.Context, project
...
+
+	endpoint := fmt.Sprintf("https://compute.googleapis.com/compute/v1/projects/%s/regions", projectId)
+	req, err := c.CreateGCPHttpRequest(ctx, "GET", endpoint, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create GCP HTTP request: %w", err)
+	}
+
+	client := httpclient.NewJsonResponseHttpClient[listRegionRespnse](c.BaseClient)
+	response, httpResponse, err := client.DoWithContext(ctx, req)
+	if httpResponse != nil && httpResponse.Body != nil {
```
jyane, line 581: I wonder this condition is actually needed: https://github.com/GoogleCloudPlatform/khi/blob/35ec20ab446b2a910e53521718a31106def28728/pkg/common/httpclient/json_response_client.go#L43

We can close the body after error check?

pkg/source/gcp/task/cloud-composer/form/autocomplete.go
RyuSA: you right, I missed it. Let me fix it.

pkg/source/gcp/api/gcp_client.go
RyuSA: I re-used that part from other implementations(i.e https://github.com/GoogleCloudPlatform/khi/blob/main/pkg/source/gcp/api/pageclient.go#L53-L55)

It sounds reasonable. If `httpResponse.Body` is really nil, `JSONReponseHttpClient.DoWithContext` must be failed. 

as I did offline communication, it seems this implementation came from the golint.

```
% podman run --rm -v $(pwd):/app -w /app golangci/golangci-lint:v1.64 golangci-lint run --config=.golangci.yaml                                 2025-04-21T12:33:27+0900
pkg/source/gcp/api/gcp_client.go:536:44: response body must be closed (bodyclose)
                        response, _, err := client.DoWithContext(ctx, req)
                                                                ^
```

let me add a TODO comment for it. We will update the implementation.

jyane: I see, I didn't notice the linter. Thanks for your explanation.

kyasbal (approved):
LGTM

PR #140, opened 2025-04-15
Prevent possible fd leak

jyane:
I think the leak should be an edge case, but adding defer to make sure that never happen just in case.


kyasbal (approved):
LGTM

PR #143, opened 2025-04-16
Fix suggested file name header field was not assigned

kyasbal:
Defined a task to supply the suggested file name in the header metadata. This is used as the default value of the file name when users click the download button on the list. Use `unnamed.khi` as its default value, and use cluster name and times for GCP clusters.


pkg/source/gcp/task/gcpcommon/filename.go
```diff
@@ -0,0 +1,58 @@
...
+	endTime := task.GetTaskResult(ctx, gcp_task.InputEndTimeTaskID.GetTaskReference())
+	startTime := task.GetTaskResult(ctx, gcp_task.InputStartTimeTaskID.GetTaskReference())
+
+	header.SuggestedFileName = getSuggestedFileName(clusterName, startTime, endTime)
+
+	return struct{}{}, nil
+}, inspection_task.NewRequiredTaskLabel(), inspection_task.InspectionTypeLabel(inspectiontype.GCPK8sClusterInspectionTypes...))
+
+func getSuggestedFileName(clusterName string, startTime, endTime time.Time) string {
+	return fmt.Sprintf("%s-%s-%s.khi", clusterName, endTime.Format("01021504"), startTime.Format("01021504"))
```
jyane, line 57: nits: Why here drops year data? The meaning of this suffix not obvious at first glance, the users likely have no idea what these numbers mean.

kyasbal: Ah I noticed this comment after the merge.

I will include the year then.

PR #147, opened 2025-04-18
Fixed issues on task graph resolution

kyasbal:
* Fix infinite loop happen when same dependency IDs are supplied in the dependencies
* Improve the cyclic dependency detection and its error message

### DIFF
--- a/pkg/task/taskset.go
+++ b/pkg/task/taskset.go
@@ -40,8 +40,8 @@ type SortTaskGraphResult struct {
 	// MissingDependencies is the list of task reference Ids missed to resolve task dependencies.
 	// This must be empty array when the sorting succeeded.
 	MissingDependencies []taskid.UntypedTaskReference
-	// HasCyclicDependency indicates if the task graph containing any cyclic dependencies.
-	HasCyclicDependency bool
+	// CyclicDependencyPath is the path of task dependencies. Runnable became  false if this field is present.
+	CyclicDependencyPath string
 	// Runnable indicate if this task graph is runnable or not. It means the tasks are sorted in topoligical order and all of input dependencies are resolved.
 	Runnable bool
 }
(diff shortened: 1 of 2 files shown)


pkg/task/taskset.go
```diff
@@ -209,46 +206,81 @@ func (s *TaskSet) sortTaskGraph() *SortTaskGraphResult {
...
-					delete(missingTaskIdsInMap, task.UntypedID().ReferenceIDString())
-				}
-			}
-
 			missingSources := []taskid.UntypedTaskReference{}
 			for source := range missingTaskIdsInMap {
 				missingSources = append(missingSources, taskid.NewTaskReference[any](source))
 			}
 
+			if len(missingSources) == 0 {
```
jyane, line 222: I guess this part is actually detecting the cyclic and building the path, however, this is complicated enough.

Can we define this part as another function so that we can focus 'sort' part?

pkg/task/taskset.go
```diff
@@ -209,46 +206,81 @@ func (s *TaskSet) sortTaskGraph() *SortTaskGraphResult {
...
+										queue[taskID] = struct{}{}
+										break
+									}
+								}
+							}
+
+						}
+					}
+				}
+				panic("unreachable")
```
jyane, line 277: I'm not sure here should panic without any logs.

It might be better to leave some hints if the logic reached here or returning error to handle this situation?

pkg/task/taskset.go
```diff
@@ -40,8 +40,8 @@ type SortTaskGraphResult struct {
 	// MissingDependencies is the list of task reference Ids missed to resolve task dependencies.
 	// This must be empty array when the sorting succeeded.
 	MissingDependencies []taskid.UntypedTaskReference
-	// HasCyclicDependency indicates if the task graph containing any cyclic dependencies.
-	HasCyclicDependency bool
+	// CyclicDependencyPath is the path of task dependencies. Runnable became  false if this field is present.
```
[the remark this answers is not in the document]
kyasbal: Updated the comment. 
But I didn't implemented these methods because I changed the type to be package private. (renamed to `sortTaskGraphResult`)
We can put the premise that users of the type are aware of the usage of these fields.

pkg/task/taskset.go
kyasbal: Done.

pkg/task/taskset.go
kyasbal: Added comment and changed the panic message to contain more message.

jyane: I'm sorry I meant it would be better to dump data for debugging if it happened?

So we can gather the debugging information from GitHub issues or from internal users/systems?

kyasbal: I added some information on there. 
But I'm not sure if it's enough because I can't imagine the actual situation to reach there.

jyane (approved):
LGTM

PR #148, opened 2025-04-18
Fix compilation error remained after the merge

kyasbal:
We seems to fail merging #141 and #143. The compilation error is remained in the main branch.


jyane (approved):

RyuSA (approved):
LGTM

PR #150, opened 2025-04-20
Support default value updated dynamically

kyasbal:
* Update the parameter store to update the parameter when backend upadates the form default value dynamically. It updates the parameter only when the parameter was the previous default value.

This is needed for supporting log view support. The form default value can change when user modified the project ID field.

### DIFF
--- a/web/src/app/dialogs/new-inspection/components/service/parameter-store.ts
+++ b/web/src/app/dialogs/new-inspection/components/service/parameter-store.ts
@@ -24,6 +24,7 @@ import {
   Subject,
   take,
   takeUntil,
+  withLatestFrom,
 } from 'rxjs';
 
 /**
@@ -42,6 +43,12 @@ export interface ParameterStore {
    * When the specified id is not available parameter for now, it waits the value to be available.
    */
   watch<T>(id: string): Observable<T>;
+
+  /**
+   * Get the observable to monitor the dirtiness of the field.
+   * true is emitted after user modified the field.
+   */
+  watchDirty(id: string): Observable<boolean>;
 
   /**
    * Watch the all parameters.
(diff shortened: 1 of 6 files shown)


web/src/app/dialogs/new-inspection/components/service/parameter-store.ts
```diff
@@ -100,12 +102,31 @@ export class DefaultParameterStore implements ParameterStore {
...
+   * Update the default values for parameters.
+   * If the current value is same as the previous default values, the parameter is updated with the newer default values.
+   * If not, the parameter value is kept because it was updated by the user.
+   */
   setDefaultValues(defaultValues: { [id: string]: unknown }): void {
     this.currentParameters
       .pipe(takeUntil(this.destroyed), take(1))
       .subscribe((parameters) => {
-        this.currentParameters.next({ ...defaultValues, ...parameters });
+        const wasDefault: { [id: string]: boolean } = {};
```
renamoo, line 114: From the perspective of design, letting the data object which represents the state of the form have a property like `dirty` in `AbstractControl` (https://angular.dev/api/forms/AbstractControl#dirty) and use them as a condition might be cleaner. 
At the same time, that require more change and I am certainly not expert on this frontend architecture so the above approach could be imfeasible in the first place. So you can disregard it if it's not applicable. Just dropping the thought.

renamoo (approved):
LGTM

web/src/app/dialogs/new-inspection/components/service/parameter-store.ts
kyasbal: I added the dirty observable. And it's also needed to show the status of the field on the form field headers.

Issue #82, opened 2025-02-25
Select Cloud Logging views or scopes


michaelbannister:
**Is your feature request related to a problem? Please describe.**

My team operates large multi-tenant clusters where container logs & events are routed to logging buckets in many tenant projects (as described in the GKE guide [Set up multi-tenant logging](https://cloud.google.com/kubernetes-engine/docs/how-to/multi-tenant-logging)). Logs _not_ related to a specific tenant go into a log bucket in the same project as the cluster (not actually the project's _Default bucket as we had to change the region).

Logging queries against the project (without specifying a scope or a set of views) result in the error "Request exceeds maximum of 200 views".

**Describe the solution you'd like**

Some way to specify a [log scope](https://cloud.google.com/logging/docs/log-scope/create-and-manage) or set of [log views](https://cloud.google.com/logging/docs/logs-views) so that I can get the queries to actually succeed.

kyasbal:
@michaelbannister Thank you for reporting this! I'd love to implement supporting this.
This feature is also important for users forwarding their logs to the other project. We need to add flexibility on the log scopes used in the steps to gather logs.

I'm currently working on #72 and it includes major improvement on form fields to generate query.
I'd like to add some forms and the functionality after we implement #72.

kyasbal:
Our API seems not to provide feature to list logs with log scopes. But.we can add functionality to gather logs with setting set of log views. 

I will work on this in this week.

PR #156, opened 2025-04-22
Epic/log views support resource names

kyasbal:
#82

### DIFF
--- a/web/src/app/dialogs/new-inspection/components/group-parameter.component.ts
+++ b/web/src/app/dialogs/new-inspection/components/group-parameter.component.ts
@@ -104,6 +104,6 @@ export class GroupParameterComponent {
    * Toggle the collapsed status for children.
    */
   toggle() {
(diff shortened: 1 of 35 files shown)


web/src/app/dialogs/new-inspection/components/group-parameter.component.ts
```diff
@@ -104,6 +104,6 @@ export class GroupParameterComponent {
    * Toggle the collapsed status for children.
    */
   toggle() {
-    this.collapsedFromUserInput.set(this.childrenStatus() === 'collapsed');
+    this.collapsedFromUserInput.set(this.childrenStatus() === 'expanded');
```
renamoo, line 107: We might want to add comment for `collapsedFromUserInput` to clarify when it's true. 
I initially thought it's true when collapsed but not sure what `FromUserInput` indicates and it seems we changes the behavior here.

renamoo (approved):
LGTM for frontend

Issue #119, opened 2025-04-01
Mixed usage of term `taskID`


kyasbal:
The server code uses `taskID` as the meaning of `inspectionID` in the inspection task runner.

https://github.com/GoogleCloudPlatform/khi/blob/main/pkg/server/server.go#L144

Standardise the meaning of these IDs in the following rule:

* `inspectionID`: An id assigned for each task graph created for `New inspection` click on frontend
* `taskID` : The ID tied with `task.Task`.
* `runID` : An unique id generated for each calls of run in a graph.

PR #157, opened 2025-04-23
Fix the wrong terminology `taskID` in APIs

kyasbal:
This is related to #119.

* The `taskID` used in server code was `inspectionID` in the task graph terminology. Use `inspectionID` instead of `taskID` in server related codes to reduce the confusion.

* Simplify API paths
  * Bumped API version from /api/v2 => /api/v3
  * /api/v2/inspection/tasks/:taskID -> /api/v3/inspection/:inspectionID
  * /api/v2/inspection/tasks -> /api/v3/inspection

### DIFF
--- a/cmd/kubernetes-history-inspector/main.go
+++ b/cmd/kubernetes-history-inspector/main.go
@@ -206,7 +206,7 @@ func run() int {
 				exitCh <- 1
 				return
 			}
-			taskId, err := inspectionServer.CreateInspection(*parameters.Job.InspectionType)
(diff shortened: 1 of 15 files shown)


jyane:
FYI: Frontend coverage CI is failing.

jyane (approved):
LGTM

renamoo (approved):
I am not sure how often we expect to update API version, but if we make it to a variable, that might reduce the cost to update everywhere in the future.

LGTM for frontend

PR #158, opened 2025-04-24
Make TaskImplementationID not to implement TaskReference

kyasbal:
TaskImplementationID interface had implemented TaskReference previously to select a dependent tasks specifically not by a task reference. This feature make users to specify dependent tasks with FQDN and unintentionally makes strictly bound tasks. Task dependencies now should all defined with TaskReference instead.

Users would need to call `GetTaskReference` more with this change. I changed the method name from `GetTaskReference()` to `Ref()` instead regarding the count of this method usage.


jyane (approved):
LGTM

kyasbal:
Resolved conflicts.
It revoked the previous review. PTAL @jyane

kyasbal:
> Resolved conflicts.
> It revoked the previous review. PTAL @jyane 

Ah, a check is failing. I'll fix first

jyane (approved):

PR #160, opened 2025-04-25
Enable parallel download for inspection data

### LINKED ISSUE #127: Download KHI file in parallel from the frontend
Currently KHI frontend receives KHI file data with splitting them at most 16MB.
KHI currently waits the response for every time but this can be done in paralllel.

We need to improve this code.
https://github.com/GoogleCloudPlatform/khi/blob/main/web/src/app/services/api/backend-api.service.ts#L184-L244

jyane:
This will resolve #127.


jyane:
Rebased to master.

jyane:
We had offline discussions and realized that some our internal environments may drop the Content-Length header so we will use metadata API instead.

I will revert [the HEAD commit](https://github.com/GoogleCloudPlatform/khi/commit/35ec20ab446b2a910e53521718a31106def28728), and will revise this PR after the revert.

jyane:
@kyasbal @renamoo The client now uses metadata API, PTAL.

web/src/app/services/data-loader.service.ts
```diff
@@ -393,37 +393,40 @@ export class InspectionDataLoaderService {
...
+            acc += done;
+            this.progress.updateProgress({
+              message: `Downloading inspection data...(${ProgressUtil.formatPogressMessageByBytes(
+                acc,
+                allSize,
+              )})`,
+              percent: (acc / allSize) * 100,
+              mode: 'determinate',
+            });
+          })(0),
```
kyasbal, line 417: Accumulating done bytes in ProgressReporter would be better to have done in `getInspectionData`?
I think the `DownloadProgressReporter` should naturally be called with the done bytes of entire file instead of a chunk sizes used in the func because the parallel behavior is hidden in the func.

kyasbal (approved):

jyane:
@kyasbal PTAL.

kyasbal (approved):
LGTM

PR #168, opened 2025-04-30
Adding functionalities used in previous structured data structure in the newer structured data structure

kyasbal:
This will be active after #167 is merged.

Adding the following features:

* Reader type provides accessible interfaces for parsers to read fields with field paths.
* Serializer related features to serialize part of structured data
  (With keeping order of keys in map in the serialized result)
* Merger related features to support stategic-merge patch requests
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md

FYI: merger structure

```mermaid
flowchart TD
	MergeNode --> mergeNode
	mergeNode -->|when the node is scalar| mergeScalarNode
	mergeNode -->|when the node is sequence| mergeSequenceNode
	mergeNode -->|when the node is map| mergeMapNode
	mergeSequenceNode -->|when the sequence items are scalar| mergeScalarSequenceNode
	mergeSequenceNode -->|when the sequence items are sequence| mergeSequenceSequenceNode
	mergeSequenceNode -->|when the sequence items are map| mergeMapSequenceNode
	mergeMapSequenceNode -->|when the patch policy is replace| mergeMapSequenceNodeWithReplaceStrategy
	mergeMapSequenceNode -->|when the patch policy is merge| mergeMapSequenceNodeWithMergeStrategy

	mergeMapNode o-..->|for each items| mergeNode
	mergeSequenceSequenceNode o-..->|for each sequences| mergeNode
	mergeMapSequenceNodeWithReplaceStrategy o-..->|for each maps| mergeNode
	mergeMapSequenceNodeWithMergeStrategy o-.->|for each maps| mergeNode
```


pkg/common/structurev2/standard.go
```diff
@@ -78,3 +197,127 @@ func (n *StandardMapNode) Children() NodeChildrenIterator {
...
+	standardRootNode, err := cloneStandardNodeFromNode(node)
+	if err != nil {
+		return nil, err
+	}
+	return standardRootNode.(json.Marshaler), nil
+}
+
+// cloneStandardNodeFromNode clones Node interface into Standard***Node.
+// Standard**Node implements json.Marshaller and yaml.Marshaller, it allows serializing any implementing Node interface.
+func cloneStandardNodeFromNode(node Node) (Node, error) {
```
jyane, line 284: I believe this part is the most important to handle large objects, but I don't see any issue so far.

jyane (approved):
I feel this PR has enough tests but I may oversight some edge cases, like empty.
Also, I was not able to track how config will be updated through the recurrsive.

@RyuSA I would recommend to get your approve as well.

PR #170, opened 2025-05-01
fix the logic error not to expand its children on click the expander

kyasbal:


renamoo (approved):
LGTM as long as my understanding that `toggle()` should return the state after toggled.

Notes: 
- true -> collapsed
- null -> unset (open)
- false -> open

- Called when it's open -> toggle returns  ("open" !== 'collapsed') meaning true, meaning it's now collapsed
- Called when it's hidden -> toggle returns  ("collapsed" !== 'collapsed') meaning false, meaning it's now open

PR #189, opened 2025-05-27
bugfix: Array merge fails when merge key item is non string type

kyasbal:
This issue would be reproducible with the following manifest on the current main branch version.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: prestop-sleep-pod-with-ports
spec:
  containers:
  - name: main-container
    image: busybox
    command: ["/bin/sh", "-c", "echo 'Container is running, waiting for termination signal.'; sleep 3600"]
    ports:
    - name: http
      containerPort: 80
      protocol: TCP
    - name: https
      containerPort: 443
      protocol: TCP
    - name: custom-port
      containerPort: 8080
      protocol: TCP
    lifecycle:
      preStop:
        exec:
          command: ["/bin/sh","-c","echo 'PreStop hook is triggered, sleeping for 30 seconds...'; sleep 30; echo 'Slept for 30 seconds, now terminating.'"]
  terminationGracePeriodSeconds: 45
```

The container ports are the key of the array. 
https://github.com/kubernetes/api/blob/v0.32.1/core/v1/types.go#L2806

### DIFF
--- a/pkg/common/structurev2/mergeoption.go
+++ b/pkg/common/structurev2/mergeoption.go
@@ -20,6 +20,26 @@ import (
 	"github.com/GoogleCloudPlatform/khi/pkg/log/structure/merger"
 )
 
+// keyItem represents a node used as a key for merge or patch merge directives.
+// It must be hashable as a string value, but also needs to retain its type.
+type keyItem struct {
+	StringKey    string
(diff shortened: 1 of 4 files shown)


RyuSA (approved):
LGTM

PR #204, opened 2025-06-19
Sass to scss

jyane:
## Summary

This PR converts all sass to scss. I'm going to enable the lint for scss after this PR.

## Points to review

Since it's impossible to review this PR by human, I performed following steps to convert the styles:

1. Downgraded dart-sass to old-fashioned ruby-sass to apply `sass-convert`
2. Applied `sass-convert` with some tweaks (mostly for downgrading the syntax) [1]
3. Reverted the tweaks manually (mostly for upgrading the syntax)

The steps are corresponding to the commits (1 2 = rename modified, 3 = fixed) in this branch, you can squash all commits while merging this PR.

Please focus on the commits separately, but only the last commit was made by manual.

## How to check the changes were valid?

Please modify angular.json and build the files (this PR vs main) as dev without sourcemap like [2], the file hash should be the same.
Also, you can confirm the file hash should be different if you modified any "scss" file.

```
[~/workspace/src/github.com/jyane/khi] (skamimori@skamimori 2025-06-19 23:38:22)
$ md5sum ~/workspace/tmp/khi/main.main.js
d900d8e6a2a0f1bb4be8e4390c5838cd  /home/skamimori/workspace/tmp/khi/main.main.js
[~/workspace/src/github.com/jyane/khi] (skamimori@skamimori 2025-06-19 23:38:27)
$ md5sum ~/workspace/tmp/khi/sass-to-scss.main.js
d900d8e6a2a0f1bb4be8e4390c5838cd  /home/skamimori/workspace/tmp/khi/sass-to-scss.main.js
```

## Reference
[1]
``` bash
#!/bin/bash

# Using bash only features

shopt -s expand_aliases
alias sass-convert="/home/skamimori/.local/share/gem/ruby/3.1.0/bin/sass-convert"

for v in $(git ls-files '*.sass');
do
  cat $v | sed \
    -e 's/@use/@import/g' \
    -e 's/mat\.//g' \
    -e 's/gen\.//g' \
    -e 's/color\.//g' \
    -e 's/common\.//g' \
    -e 's/cp\.//g' \
    -e 's/map\.//g' \
    -e 's/:host::ng-deep/dummy-selector/g' \
  rm $v
  sass-convert -F sass -T scss tmp > $v
  git mv $v "${v%.sass}.scss"
done

rm tmp
```

[2]
``` diff
diff --git a/scripts/generate-angular-json.sh b/scripts/generate-angular-json.sh
index 27f8973..124bf39 100755
[...]

PR #206, opened 2025-06-20
Re-enable scss lint

jyane:
The lint was disabled in #186.

Since #204 was merged into main, I would like to re-enable this lint for scss.


kyasbal (approved):
LGTM 🥳

PR #212, opened 2025-06-23
fix(ci): Updated Node.js version on CD pipeline from 18 to 22

kyasbal:
This fixes the failing deployment pipeline after upgrading Angular version.


RyuSA (approved):
LGTM

PR #220, opened 2025-07-01
Convert generated.sass to scss

jyane:
All sass files were supposed to be eliminated in #204, however, the file generated.sass was left because it is not git tracked.


jyane:
Sorry, I think I have to update some parts as well. Let me see.

kyasbal (approved):

PR #228, opened 2025-07-09
Ensure token not to be nil while the error not nil

jyane:
This issue is found in our internal flaky tests.
The flakiness is easily reproducible if you increase the number of executions.
Please see also b/429906681 (only for internal).

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x588edc]

goroutine 186011 [running]:
github.com/GoogleCloudPlatform/khi/pkg/common/token.TestBasicTokenStore_GetTokenCallsResolverOnceInParallel.func1.1()
    /workspace/pkg/common/token/store_test.go:110 +0x9c
created by github.com/GoogleCloudPlatform/khi/pkg/common/token.TestBasicTokenStore_GetTokenCallsResolverOnceInParallel.func1 in goroutine 119455
    /workspace/pkg/common/token/store_test.go:103 +0x1b3
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x588edc]
```


pkg/common/token/store.go
```diff
@@ -59,6 +60,12 @@ func (b *BasicTokenStore) GetType() string {
 
 // GetToken implements TokenStore.
 func (b *BasicTokenStore) GetToken(ctx context.Context) (*Token, error) {
+	// This ensures lastToken not to be nil while b.lastTokenRefreshError != nil.
```
kyasbal, line 63:
```suggestion
	// This ensures lastToken not to be nil while b.lastTokenRefreshError == nil.
```
Or maybe my understanding is wrong

kyasbal (changes requested):

kyasbal (approved):

PR #229, opened 2025-07-09
Simplify GetToken

jyane:
I have realized that the if statement is likely intended to initialize the token for the first call.
I think we can simplify it since the initialization was added in #228.

This could have a race condition if RefreshToken is called while initializing the token, however, it should be a rare case and supposed to be a wrong use case.

### DIFF
--- a/pkg/common/token/store.go
+++ b/pkg/common/token/store.go
@@ -61,18 +61,13 @@ func (b *BasicTokenStore) GetType() string {
 // GetToken implements TokenStore.
 func (b *BasicTokenStore) GetToken(ctx context.Context) (*Token, error) {
 	// This ensures lastToken not to be nil while b.lastTokenRefreshError == nil.
-	// Also, sync.Once guarantees there is only 1 execution across all goroutines.
-	// Thus, we can set the token without the lock.
+	// Also, sync.Once.Do guarantees there is only 1 execution across all goroutines, and it blocks others until ends.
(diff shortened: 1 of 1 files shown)


pkg/common/token/store.go
```diff
@@ -61,18 +61,13 @@ func (b *BasicTokenStore) GetType() string {
...
 	// This ensures lastToken not to be nil while b.lastTokenRefreshError == nil.
-	// Also, sync.Once guarantees there is only 1 execution across all goroutines.
-	// Thus, we can set the token without the lock.
+	// Also, sync.Once.Do guarantees there is only 1 execution across all goroutines, and it blocks others until ends.
+	// Thus, we can set the token without the lock and can assume either lastToken or lastTokenRefreshError should be set.
 	b.initialRefreshOnce.Do((func() {
 		b.refreshTokenWithoutLock(ctx)
 	}))
 	defer b.tokenLock.RUnlock()
 	b.tokenLock.RLock()
```
kyasbal, line 70: maybe this is also not needed?

kyasbal (changes requested):

pkg/common/token/store.go
jyane: Interesting, it's probably true.

Readers may see either an old token (someone already verified it's actually outdated) or an old error but this shouldn't be an issue?

kyasbal: OK. then let's keep it.

PR #232, opened 2025-07-15
style: fix an issue that prevents inspection from scrolling

jyane:
This was probably unintentionally commited in #202.

@kyasbal This is preventing the new inspection screen from scrolling, then the inspection became not usable.
However, I'm not sure why it was changed, so please take a look at this again.

Thank you!


kyasbal:
@jyane Isn't this scrollable?
image

PR #248, opened 2025-08-12
Simplify log metadata and moved log logics under inspection/logger

kyasbal:
Current metadata logger implementation contains the type works as a metadata of a task and a type used as a `slog.Handler`. For the later refactoring of metadata package, we'll remove dependency from the `metadata/logger` as much as possible. 
In this change, I refactored the log metadata and now it just contains `bytes.Buffer` log buffer and it's written by `inspection/logger`. This change also splits the TaskLogger implementations to multiple single role slog.Handlers rather than the previous monolisic implementation.


pkg/inspection/logger/severity_filter_test.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"bytes"
+	"context"
+	"log/slog"
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestSeverityFilter(t *testing.T) {
```
RyuSA, line 26: Could you add extra tests for `WithAttrs` and `WithGroup`? Since this is a core package I would add detailed tests.

pkg/inspection/logger/throttle_filter_test.go
```diff
@@ -0,0 +1,99 @@
...
+package logger
+
+import (
+	"bytes"
+	"log/slog"
+	"strings"
+	"testing"
+)
+
+func TestThrottleFilter(t *testing.T) {
```
RyuSA, line 24: Could you add extra tests for WithAttrs and WithGroup? Since this is a core package I would add detailed tests.

pkg/core/inspection/runner.go
```diff
@@ -441,7 +478,18 @@ func (i *InspectionTaskRunner) addCommonMetadata(ctx context.Context, writableMe
 	}
 	typedmap.Set(writableMetadata, plan.InspectionPlanMetadataKey, plan.NewInspectionPlan(taskGraphStr))
 
-	i.MakeLoggers(ctx, getLogLevel(), writableMetadata.AsReadonly(), taskGraph.GetAll())
+	logMetadata := i.makeLoggers(ctx, getLogLevel(), taskGraph.GetAll())
+	typedmap.Set(writableMetadata, log.LogMetadataKey, logMetadata)
+}
+
+func (i *InspectionTaskRunner) cleanupAfterAnyRun(ctx context.Context, taskGraph *coretask.TaskSet) {
```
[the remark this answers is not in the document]
kyasbal: The logger itself is called from the global logger.
https://github.com/GoogleCloudPlatform/khi/pull/248/files/68059a4302e3e3eea9d4f5b3fa74bd603d2930c7#diff-e2927a1cfdd9e667a6dd21ae5de78b3750f6f85ee96dd0f0d5b4c87ad29d24c2R97

It manages a map from `task ID`,`run ID` and `Inspection ID` to a logger created for a logger. The map reference itself consumes low memory, that's fine. But it eventually reference the buffer used in the log metadata.
After finishing a task to run, the global logger doesn't need to have the reference to the task logger to avoid memory leak. The log buffer could be large.

Indeed this clean-up wasn't handled correctly in the past implementation.

pkg/inspection/logger/throttle_filter_test.go
kyasbal: Added

RyuSA (approved):

PR #257, opened 2025-08-21
Move source/gcp tasks under task/inspection sub folders

kyasbal:


kyasbal:
These commits will be another multiple PRs.

PR #262, opened 2025-08-28
Remove the role of task resolution from TaskSet and add another extensible task resolver type

kyasbal:
# Motivation

To run user selected parser tasks with resolving its dependency, KHI adds dependencies of parser tasks and its dependencies recursively then sort tasks topological order. This 2 roles would be better to split and  the dependency resolutions would be better to be extensible although the topological sort wouldn't be expected to have extended feature from now.

# Changes

* This change introduce a new type called `GraphResolver` and an interface to define task resolution rule `GraphResolverRule`. 
* Added new implementation of GraphResolverRule
  * RequiredTaskLabelGraphResolverRule: Include the result task list always if a task has required label
  * TaskDependencyGraphResolverRule: Include tasks depended from the other task already included.
  * SubsequentTaskRefsGraphResolverRule: Reverse order depenency. It read a specific task label and add tasks in the result and added tasks have the task dependency on the task having the label.
     * This is used for example, a parser tasks must be followed by the serializer task. In the case, serializer tasks are automatically added and it must be run after the previous parser tasks completed.
     * This rule simplify dynamatically constructed task graph. WrapTask method used in TaskSet is no longer needed because of this rule addition.

### DIFF
--- a/pkg/core/inspection/legacyparser/parser.go
+++ b/pkg/core/inspection/legacyparser/parser.go
@@ -180,5 +180,6 @@ func NewParserTaskFromParser(taskId taskid.TaskImplementationID[struct{}], parse
 	},
 		append([]coretask.LabelOpt{
 			inspectioncore_contract.FeatureTaskLabel(parser.GetParserName(), parser.Description(), parser.TargetLogType(), featureOrder, isDefaultFeature, availableInspectionTypes...),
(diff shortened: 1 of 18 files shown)


RyuSA (approved):
TaskSet does not resolve task graph, LGTM.

Issue #267, opened 2025-09-08
Changes made by delete collection operation may generate wrong resource timeline


kyasbal:
Current KHI logic to handle delete collection operation can cause an error timeline. Because it has no `metadata.deletionTimestamp` field patching in its response or request field, the later changes can't detect the right deletion starting timing. 

<img width="3200" height="1661" alt="Image" src="https://github.com/user-attachments/assets/83754de6-10ef-477f-a709-67ce0121b9e1" />

PR #272, opened 2025-09-12
fix bug #267:  manifest generator to be aware of uid change

kyasbal:
Screenshot 2025-09-12 at 15 00 22

* Left: wrong old timeline. There are several red(deleted) revisions during deleting period.
* Right: fixed.

---

This change added logic to be aware of `metadata.uid`. When there is a deleteCollection request in a middle of PATCH and an actual DELETE, the previous implementation couldn't merge the resource manifest well because the delete collection request audit log doesn't have its resource body. This change reads the UID field of metadata from logs with resource body and it resets the previous resource manifest status when the uid is changed.

This will be marked as ready after #268 and #271 is reviewed and merged.

### DIFF
--- a/pkg/task/inspection/commonlogk8saudit/impl/manifestgenerate_task.go
+++ b/pkg/task/inspection/commonlogk8saudit/impl/manifestgenerate_task.go
@@ -65,8 +65,8 @@ var ManifestGenerateTask = inspectiontaskbase.NewProgressReportableInspectionTas
 	for _, group := range groups {
 		currentGroup := group
 		workerPool.Run(func() {
-			prevRevisionBody := ""
-			prevResourceUID := "unknown"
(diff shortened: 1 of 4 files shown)


pkg/task/inspection/commonlogk8saudit/impl/manifestgenerate_task.go
```diff
@@ -66,12 +66,14 @@ var ManifestGenerateTask = inspectiontaskbase.NewProgressReportableInspectionTas
 		currentGroup := group
 		workerPool.Run(func() {
 			prevRevisionBody := ""
+			prevResourceUID := "unknown"
```
jyane, line 69: I assume UID should be generated as UUID and won't be a such string.

jyane (approved):

Issue #285, opened 2025-10-03
Migrate our self implemented REST API client against Google APIs to the standard SDK


kyasbal:
Our current implementation just uses the REST API with the standard Go http client. We should aim to use the standard SDK instead of them. 

I'll update this issue to post the detailed design for this change later.

PR #295, opened 2025-10-07
[feat #285 5-1/N] Replace API client to v2 for GKE cluster name auto completion

kyasbal:
This is the series of changes to modify current auto completion logic to use the new API client. This change is only for GKE.


kkuchima (approved):
LGTM

PR #320, opened 2025-10-18
Rename API package name from `googlecloudv2` to `googlecloud`

kyasbal:
This change only contains the change to rename the folder and its package name.


kkuchima (approved):
LGTM

PR #324, opened 2025-10-20
Migrate API client from REST based self implementation to new client using official SDK

kyasbal:
This PR merge the long epic branch `epic/issue-285-api-client-v2` into `main`.

# Motivation

The API client used in KHI implementation was implemented on the basic HTTP client by ourself for historical reason. Now this become burden to maintain and it usually becomes a wall to start using another API.
This change remove these old client implementation and use the official SDKs.

# Architecture change

This PR replaces the API client defined under `pkg/api/googlecloud`. Now all the clients must be created from the ClientFactory defined in this package. 
All KHI tasks must not use the client directly regarding the testability. With replacing the API client in this PR, each task packages may have injection task to supply implementation for the interface to actually calling the API. This make us easier to test tasks calling API just by providing mock API client for the specific task.

API client users in task should use the following tasks to obtain the client:

* `googlecloudcommon_contract.APIClientFactoryTaskID`: The task to inject the instance of ClientFactory
* `APIClientCallOptionsInjectorTaskID`: The task to supply the CallOptionInjector which provides options for per RPC calls.

For the references of using the new client architecture in KHI task system, please refer implementation of `pkg/task/inspection/googlecloudcommon_impl/logfetcher.go` which is the actual implementation calling listLogEntries API on Cloud Logging.

### DIFF
--- a/pkg/api/googlecloud/legacy/rawtoken.go
+++ b/pkg/api/googlecloud/legacy/rawtoken.go
@@ -1,0 +1,41 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


RyuSA (approved):
LGTM

PR #330, opened 2025-10-28
bug: fix wrong validation error can block user to run inspection on slow network environment

kyasbal:
Users with slow internet connection couldn't press the Run button because the dryrun requests were unintentionally cancelled by switchMap.

### DIFF
--- a/web/src/app/dialogs/new-inspection/new-inspection.component.ts
+++ b/web/src/app/dialogs/new-inspection/new-inspection.component.ts
@@ -159,7 +159,7 @@ export class NewInspectionDialogComponent implements OnDestroy {
         client.dryrun(req);
       });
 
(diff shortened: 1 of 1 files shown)


web/src/app/dialogs/new-inspection/new-inspection.component.ts
```diff
@@ -158,9 +159,13 @@ export class NewInspectionDialogComponent implements OnDestroy {
         client.dryrun(req);
       });
 
-    // Send dryrun request to server when any of the parameters changed to validate parameters.
-    this.store
-      .watchAll()
+    // Send dryrun request to server when any of the parameters changed or every secounds to validate parameters.
```
renamoo: nit: secounds -> seconds

renamoo (approved):

PR #403, opened 2025-12-05
Updated package-lock.json & fixed several compile errors due to updates

kyasbal:


web/src/app/common/misc-util.ts
```diff
@@ -0,0 +1,43 @@
...
+ *     break;
+ *   default:
+ *     unreachable(a); // this won't be an error.
+ * }
+ *
+ * switch(a) {
+ *   case "A":
+ *     break;
+ *   default:
+ *     unreachable(a); // this will be an error because the 'a' is "B" and not never.
```
renamoo: I might not be able to understand this example well. So, First case covers all potential value so it does not reach there and that's why we don't get error. The second case fails to handle value "B" but it can be "B" so we could get an error and notice the error in logic. Is it correct?

renamoo (approved):
I mistakenly added a comment as a single comment. Anyway, I felt example description can be clarified further but it's nit and other than that, LGTM

PR #422, opened 2025-12-23
Add customizable shared input for set values

kyasbal:
Adding a new input control component for set values. This will be used in the new inspection dialog later and replaces the filter components on the tool bar.

image

The functionality is same as the set input component on the tool bar. This component allows component user to replace the chip appearance and text suggestion  appearance.


renamoo (commented):
Added a few comments!

PR #423, opened 2025-12-23
Add set input parameter form

kyasbal:
This PR add a new parameter type `set` on the new inspection dialog with utilizes the new component introduced in #422 

Several fields in the new inspection dialog accepts aliased item or subtractive item notation. The new implementation shows these characteristic effectively.

image

Waiting #422 is merged.


kyasbal:
I opened #426 instead.

PR #432, opened 2025-12-24
Improve the set type parameter form

kyasbal:
This PR is merging the epic branch `epic/add-set-input` into main.

This change adds the set type input form and replaced parameter form using text form parameter with this new set parameter form. This change also adds new autocomplete suggestion logic that is now possible to suggest to user via the new control with obtaining suggesting values from Cloud Monitoring labels.

With these enhancement, now we provides more intuitive form to receive set type input on the new inspection form.

image


renamoo (approved):
I believe the changes on this branch are separately reviewed already, so just providing LGTM

PR #449, opened 2026-01-13
Add TimelineCornerIndicatorComponent and TimelineLegendComponent

kyasbal:
Adding 2 components used in the new timeline components.

image

**TimelineCornerIndicatorComponent**
It changes to the Scaling mode when user holds shift key.
image

**TimelineLegendsComponent**

image
image

### DIFF
--- a/web/src/app/timeline/components/timeline-legend.component.html
+++ b/web/src/app/timeline/components/timeline-legend.component.html
@@ -84,7 +84,7 @@
             </div>
           }
           <p class="severity-info">
-            Upper part of each events represents its severity.
+            The upper half of each diamond represents its severity
           </p>
         }
       }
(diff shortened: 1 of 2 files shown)


web/src/app/timeline/components/timeline-legend.component.html
```diff
@@ -0,0 +1,113 @@
...
+            <div class="legend-container">
+              <div
+                class="event-rect"
+                [style.--event-color]="legend.color"
+              ></div>
+              <p>{{ legend.label }}</p>
+            </div>
+          }
+          <p class="severity-info">
+            Upper part of each events represents its severity.
```
renamoo: nit: It's just my personal impression, but I found this text not friendly enough for users.
Probably the reason is "upper" refers to the physical shape of diamond while "of each events" suddently talks about the what it represents. 

Some ideas:
- The upper half of each icon indicates event severity
- The upper part of each event icon represents its severity
- The upper half of each diamond represents its severity

renamoo (approved):
Only a nit about text so LGTM

PR #458, opened 2026-01-15
Added timeline header components

kyasbal:
This PR adds a new TimelineRulerComponent placed on the top of the timeline graph.

**TimelineRulerComponent without log filter applied**
image

**TimelineRulerComponent with log filter applied**
image


kyasbal:
> Oops, I didn't notice it is now back to draft as I did not reload 😂 Sorry!

It's totally not a problem. I was thinking about separating this PR granular if this was too big for review. But it's passed, then fine.

PR #461, opened 2026-01-16
bugfix: EndpointSliceMapper can panic with dereference nil pointer

kyasbal:
If EndpointSlice resource audit log didn't contain the manifest body, it crashes.


SecretSword2 (approved):
Something nice.

PR #476, opened 2026-01-22
fix CI failure by running unncessary pre build step

kyasbal:
The `generate-backend` target was marked as `.PHONY` unnecessary. It required a building step on release CI script runs go command even the container image had no go image. This was triggered by the Makefile misconfiguration and this PR fixes it.


kyasbal:
I'll close this but this issue is addressed in #488.

Issue #121, opened 2025-04-02
Support GKE/Anthos clusters having exactly same name cluster in another location


kyasbal:
### Overview

Current KHI is not care the situation that 2 or more clusters created as a same name but with different location.
We need to add forms and improve the logging query to support them.

PR #494, opened 2026-02-01
Enhance GKE cluster identification by making Location field mandatory and filtering queries by location

kyasbal:
## Description

Previously, the `Location` field was only mandatory for Cloud Composer inspections. For GKE inspections, clusters were primarily identified by `Project ID` and `Cluster Name`. This caused an issue where logs from different clusters could be intermingled if multiple clusters with the same name existed across different locations within the same project.

This Pull Request addresses this issue by extending the `Location` field requirement to GKE inspections and ensuring all log queries are filtered by location.

### Key Changes

* **Mandatory Location for GKE**: Updated the GKE inspection form to require the `Location` field, consistent with Cloud Composer.
* **Improved Location Suggestion**: Enhanced the auto-complete logic. When a cluster name is entered, the system now queries Cloud Monitoring to fetch only the specific locations where a cluster of that name actually exists.
* **Refined Log Queries**: Updated all relevant log retrieval tasks to include `location` in their filter expressions. This eliminates the possibility of log collision between clusters sharing the same name in different regions or zones.

## Expected Behavior

* Users must specify a `Location` when setting up a new GKE inspection.
* The `Location` suggestion dropdown should only display relevant locations based on the entered cluster name.
* Inspection results should only contain logs from the specific cluster located in the selected project, name, and location.

### DIFF
--- a/web/src/app/timeline/components/misc/demo-builder.ts
+++ b/web/src/app/timeline/components/misc/demo-builder.ts
@@ -73,7 +73,11 @@ export class DemoViewModelBuilder {
     endTime: number,
     revisionState: RevisionState,
     verb: RevisionVerb,
+    logTime: number = NaN,
   ) {
+    if (Number.isNaN(logTime)) {
+      logTime = startTime;
+    }
     const logIndex = this.logIndex++;
     this.logs.push(
       new LogEntry(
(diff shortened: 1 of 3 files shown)


kyasbal:
This will fix #121

kkuchima (approved):

PR #495, opened 2026-02-01
Fix several styling issues

kyasbal:
image

* Changed fonts to "Google Sans Code" in the diff view and log view.
* Used smaller fonts in some controls where it usually not important or its important to see as many as possible.
* Fixed TimelineIndex component's wrong shrinking behavior
* Fixed z-index order issue of the timeline hover component (The hover component was behind of TimelineIndex or TimelineRuler)
* Adjusted margins

### DIFF
--- a/web/src/app/diff/diff-view.component.ts
+++ b/web/src/app/diff/diff-view.component.ts
@@ -172,8 +172,6 @@ export class DiffViewComponent implements OnInit, OnDestroy {
 
   public $logs = this._inspectionDataStore.allLogs;
 
-  revisionPaneHeight: BehaviorSubject<number> = new BehaviorSubject(100);
-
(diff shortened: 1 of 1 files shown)


web/src/app/diff/diff-view.component.html
```diff
@@ -23,139 +23,155 @@
...
+                    </ngx-unified-diff>
+                  } @else {
+                    <pre
+                      class="log-body-container"
+                    ><code [highlight]="diffViewModelValue.currentRevision.resourceContent" language="yaml" lineNumbers></code></pre>
+                  }
+                } @else {
+                  <ngx-unified-diff
+                    class="ngx-diff-light-theme"
+                    before=""
```
renamoo: Is this empty property intended?

PR #512, opened 2026-02-19
Optimized memory allocation for WebGL renderers

kyasbal:
KHI created ArrayBuffers for each WebGL buffer resources even the ArrayBuffer is only used for the initialization time. This frequent memory allocation made memory pressure on browser and it may finally results in the error `Invalid array length` for the environment with shortage of memory.

This PR reuses ArrayBuffer when the buffer is not necessary to be updated after its initialization.

### DIFF
--- a/web/src/app/timeline/components/canvas/glcontextmanager.spec.ts
+++ b/web/src/app/timeline/components/canvas/glcontextmanager.spec.ts
@@ -32,7 +32,7 @@ describe('GLContextManager', () => {
       'render',
       'resize',
     ]);
-    manager = new GLContextManager(canvas, renderer);
+    manager = new GLContextManager(canvas, renderer, () => {});
   });
 
   afterEach(() => {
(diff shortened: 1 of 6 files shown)


web/src/app/timeline/components/canvas/glutil.ts
```diff
@@ -204,3 +204,71 @@ export class WebGLUtil {
...
+   */
+  dataView(sizeInBytes: number): DataView {
+    this.ensureSize(sizeInBytes);
+    return new DataView(this.bufferSource, 0, sizeInBytes);
+  }
+
+  private ensureSize(sizeInBytes: number): void {
+    if (this.bufferSource.byteLength < sizeInBytes) {
+      try {
+        const newSize = Math.max(sizeInBytes, this.bufferSource.byteLength * 2);
```
jyane: 👍

PR #524, opened 2026-02-26
Generalize bisect utility

kyasbal:


web/src/app/common/misc-util.ts
```diff
@@ -42,20 +42,23 @@ export function unreachable(v: never): never {
...
  * Equivalent to Python's `bisect_left`.
  * Returns the leftmost insertion point for `x` in sorted array `arr`
  * to maintain the sorted order.
  */
-export function bisectLeft(
-  arr: number[],
+export function bisectLeft<T>(
+  arr: T[],
   x: number,
+  extractor: (item: T) => number,
```
jyane: I wonder this is actually needed. I guess we may want to change the type of `x` to `T`?

PR #614, opened 2026-04-14
fix(googlecloudk8scommon): fix duplicate cluster name prefix bug

kyasbal:
InputClusterNameTask incorrectly prepended the cluster type prefix, causing duplicate prefixes like `awsClusters/awsClusters/cluster-name` when combined with other components. This change ensures InputClusterNameTask returns only the raw cluster name.

- Remove `ClusterNamePrefixTaskRef` dependency from `InputClusterNameTask`.
- Modify `InputClusterNameTask` to use raw `ClusterName` instead of `NameWithClusterTypePrefix` in `WithHintFunc` to avoid false warnings.
- Extract inline validator and converter to package-level functions (`validateClusterName`, `convertClusterName`) for better testability.
- Update unit tests in `inputclustername_task_test.go` to cover the new functions and reflect the changed error message.


kkuchima (approved):
LGTM

PR #621, opened 2026-04-20
Adding KHI file schema v6 namespaced ID generator

kyasbal:
This PR adds new IDGenerator type which produces unique ID thread-safely. This type is used for serializing the new KHI file format v6.

### DIFF
--- a/pkg/model/khifile/v6/ids_test.go
+++ b/pkg/model/khifile/v6/ids_test.go
@@ -1,0 +1,102 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package khifilev6
+
+import (
+	"sync"
+	"testing"
+)
+
+func TestIDGenerator(t *testing.T) {
+	t.Run("sequential new", func(t *testing.T) {
+		g := &IDGenerator{}
+
+		steps := []struct {
+			name string
+			ns   IDNamespace
+			want uint32
+		}{
+			{"first string", IDString, 1},
+			{"second string", IDString, 2},
+			{"first fieldset", IDFieldSet, 1},
+		}
+
+		for _, step := range steps {
+			t.Run(step.name, func(t *testing.T) {
+				got := g.New(step.ns)
+				if got != step.want {
+					t.Errorf("New(%v) = %d, want %d", step.ns, got, step.want)
+				}
+			})
+		}
+	})
+
+	t.Run("set and new", func(t *testing.T) {
+		g := &IDGenerator{}
+		g.Set(IDString, 5)
+
+		steps := []struct {
+			name string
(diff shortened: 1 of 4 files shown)


pkg/model/khifile/v6/ids_test.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	})
+
+	t.Run("nsCounter size", func(t *testing.T) {
+		const expectedSize = 64
+		gotSize := unsafe.Sizeof(nsCounter{})
+		if gotSize != expectedSize {
+			t.Errorf("expected nsCounter size to be %d, got %d", expectedSize, gotSize)
+		}
+	})
```
jyane, line 101: This is quite interesting, so the data can fit into the CPU cache? I'm just curious but there is a trade-off between "on cache size" vs extra 60 bytes (this is actually extra 16x), is it worth investing?

jyane (approved):
left a nit comment but LGTM.

pkg/model/khifile/v6/ids_test.go
jyane: I have noticed that there is a struct in x/sys/cpu : https://pkg.go.dev/golang.org/x/sys/cpu#CacheLinePad

Probably aarch64 often uses 128 bytes for the cacheline size, so it would be better to use this?

```
// CacheLinePad is used to pad structs to avoid false sharing.
type CacheLinePad struct{ _ [cacheLineSize]byte }
```

kyasbal: Regarding the total needed ID count on KHI, it won't change a lot. But still it can be 6 times differences.
It just need to add the CacheLinePad in struct. So, I think we can just keep it.

ID namespace count will be constant by a version of KHI file. Thus the 60 byte over head is also very small. We can include this or just allow the false sharing regarding these small amount. Whichever fine for this.

```
   1 goos: linux
   2 goarch: amd64
   3 pkg: github.com/GoogleCloudPlatform/khi/pkg/model/khifile/v6
   4 cpu: AMD EPYC 7B12
   5 BenchmarkIDGenerator_Padded-24          1000000000               2.457 ns/op           0 B/op          0 allocs/op
   6 BenchmarkIDGenerator_Unpadded-24        498671058               12.00 ns/op            0 B/op          0 allocs/op
   7 PASS
   8 ok      github.com/GoogleCloudPlatform/khi/pkg/model/khifile/v6 9.924s
```

kyasbal: This is the bench code:

```go

type nsCounterUnpadded struct {
	value atomic.Uint32
}

type idGeneratorUnpadded struct {
	counters [idNamespaceMax]nsCounterUnpadded
}

func (g *idGeneratorUnpadded) New(ns IDNamespace) uint32 {
	return g.counters[ns].value.Add(1)
}

func BenchmarkIDGenerator_Padded(b *testing.B) {
	g := &IDGenerator{}
	var wg sync.WaitGroup
	wg.Add(2)
	go func() {
		defer wg.Done()
		for i := 0; i < b.N; i++ {
			g.New(IDString)
		}
	}()
	go func() {
		defer wg.Done()
		for i := 0; i < b.N; i++ {
			g.New(IDFieldSet)
		}
	}()
	wg.Wait()
}

func BenchmarkIDGenerator_Unpadded(b *testing.B) {
	g := &idGeneratorUnpadded{}
	var wg sync.WaitGroup
	wg.Add(2)
	go func() {
		defer wg.Done()
		for i := 0; i < b.N; i++ {
			g.New(IDString)
		}
	}()
	go func() {
		defer wg.Done()
		for i := 0; i < b.N; i++ {
			g.New(IDFieldSet)
		}
	}()
	wg.Wait()
}
```

PR #629, opened 2026-04-23
chore: fix make setup error in jj and update dev guide

kyasbal:
Now I'm gonna `jj` is a recommended VCS used in this project.
This PR changes following parts:
* Fixed `make setup` works without `.git` folder for the case when use `jj workspace` feature.
* Updated the developer guide related to `jj`.


scripts/generate-version.sh
```diff
@@ -20,7 +20,8 @@ PROD_ENV_NAMES=$(echo $PROD_ENV_FILES| jq -R -s -c '[split(" ")[]|capture(".web/
 DEV_ENV_NAMES=$(echo $DEV_ENV_FILES| jq -R -s -c '[split(" ")[]|capture(".web/src/environments/environment.(?<env>.*).ts")|.env]')
 
 PROD_VERSION_CONTENT=$(echo "export const VERSION=\"$(cat VERSION)\"")
-DEV_VERSION_CONTENT=$(echo "export const VERSION=\"$(cat VERSION)@$(git log -1 --pretty=format:%h )\"")
+GIT_HASH=$(git log -1 --pretty=format:%h 2>/dev/null || jj log --no-graph -T 'commit_id.short()' -r @ 2>/dev/null || echo "unknown")
```
SecretSword2: Note: `git rev-parse --short HEAD` and `git log -1 --pretty=format:%h` typically return a 7-character hash, while jj's `commit_id.short()` generally returns something longer. Are you aware of this?

```
% jj log --no-graph -T 'commit_id.short()' -r @-
e0a179e7c89c
% git log -1 --pretty=format:%h
e0a179e
```

.agents/rules/ts-coding-rule.md
```diff
@@ -11,6 +11,7 @@ globs: **/*.ts
 * Make fields `readonly` as much as possible.
 * Define enum type rather than define a union of string literal types.
 * Write import path in absolute form from the project base path like `src/app/...`.
+* Run `make format-web` and `make lint-web` to follow the standard typescript conding convention in this project.
```
SecretSword2: Typo: conding -> coding

SecretSword2 (changes requested):

scripts/generate-version.sh
```diff
@@ -20,7 +20,8 @@ PROD_ENV_NAMES=$(echo $PROD_ENV_FILES| jq -R -s -c '[split(" ")[]|capture(".web/
 DEV_ENV_NAMES=$(echo $DEV_ENV_FILES| jq -R -s -c '[split(" ")[]|capture(".web/src/environments/environment.(?<env>.*).ts")|.env]')
 
 PROD_VERSION_CONTENT=$(echo "export const VERSION=\"$(cat VERSION)\"")
-DEV_VERSION_CONTENT=$(echo "export const VERSION=\"$(cat VERSION)@$(git log -1 --pretty=format:%h )\"")
+GIT_HASH=$(git log -1 --pretty=format:%h 2>/dev/null || jj log --no-graph -T 'commit_id.short(7)' -r @ 2>/dev/null || echo "unknown")
```
SecretSword2: In both this script and the `Makefile`, the hash resolution prioritizes `git` over `jj`. 

**The Catch**: Many developers use `jj` as a wrapper around a Git backend (a colocated workspace where both `.git` and `.jj` exist). In these setups, `git log` and `git rev-parse` will *always* succeed. However, `jj` does not automatically advance the Git `HEAD` to the working commit (`@`) as you edit code. Consequently, for a `jj` developer in a colocated repo, `git log` will return the hash of the parent commit (or the checked-out branch), completely ignoring their active working commit (`@`).

**Recommendation**: If you want `jj` users to consistently get the hash of their actual working environment, we should prioritize `jj` or explicitly check for the presence of a `jj` workspace first.

```suggestion
if jj root >/dev/null 2>&1; then
  GIT_HASH=$(jj log --no-graph -T 'commit_id.short(7)' -r @ 2>/dev/null || echo "unknown")
else
  GIT_HASH=$(git log -1 --pretty=format:%h 2>/dev/null || echo "unknown")
fi
```

Makefile
```diff
@@ -1,7 +1,7 @@
 .DEFAULT_GOAL := help
 
 VERSION := $(shell cat ./VERSION)
-GIT_SHORT_HASH := $(shell git rev-parse --short HEAD)
+GIT_SHORT_HASH := $(shell git rev-parse --short HEAD 2>/dev/null || jj log --no-graph -T 'commit_id.short(7)' -r @ 2>/dev/null || echo "unknown")
```
SecretSword2: Similar to `scripts/generate-version.sh`, this will incorrectly fall back to the git `HEAD` rather than the `jj` working commit (`@`) if the developer is using a colocated `jj` + `git` workspace. We should prioritize `jj` if the user is in a jj repository.

```suggestion
GIT_SHORT_HASH := $(shell if jj root >/dev/null 2>&1; then jj log --no-graph -T 'commit_id.short(7)' -r @ 2>/dev/null; else git rev-parse --short HEAD 2>/dev/null; fi || echo "unknown")
```

scripts/make/hooks.mk
```diff
@@ -1,12 +1,23 @@
...
-	@scripts_files=$$(git diff --cached --name-only --diff-filter=ACMR | sed 's| |\\ |g'); \
-	if [ -n "$$scripts_files" ]; then \
-		$(MAKE) add-licenses && \
-		$(MAKE) format && \
-		$(MAKE) lint && \
-		$(MAKE) lint-markdown-fix && \
-		$(MAKE) lint-markdown && \
-		echo "$$scripts_files" | xargs git add; \
+	@if jj root >/dev/null 2>&1 && [ -z "$$(git diff --cached --name-only 2>/dev/null)" ]; then \
+		scripts_files=$$(jj diff --name-only | sed 's| |\\ |g'); \
```
SecretSword2: **Nit:** In the `git` implementation below, `git diff --cached --name-only --diff-filter=ACMR` deliberately excludes deleted files. However, `jj diff --name-only` does not filter out deletions. 

If a developer only deletes a file and runs `make pre-commit`, `$$scripts_files` will contain the deleted file name. This means `[ -n "$$scripts_files" ]` evaluates to true, triggering `make format` and `make lint` across the project. 

It shouldn't break anything, but it's a minor divergence in behavior. You can safely ignore this if you feel the extra formatting run is harmless!

SecretSword2 (approved):
雑にAIレビューしました。コメント見て良ければSubmitしてくだちい。

scripts/generate-version.sh
kyasbal: This is definitely a good catch. It totally makes sense.

PR #630, opened 2026-04-24
feat(web): add v6 parser binary reader and parser

kyasbal:
This PR introduces KHIFileParser and BinaryReader which handles parsing the container structure of the new KHI file.

### DIFF
--- a/web/src/app/parser/core/binary-reader.ts
+++ b/web/src/app/parser/core/binary-reader.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { assertNecessaryAPI } from 'src/app/common/misc-util';
+
 /**
  * Data associated with a single chunk in the KHI file.
  */
@@ -34,6 +36,7 @@ export class BinaryReader {
    * Creates a new BinaryReader for the given buffer.
    */
   constructor(private readonly buffer: ArrayBuffer) {
+    assertNecessaryAPI('DecompressionStream');
     this.dv = new DataView(buffer);
   }
 
@@ -55,6 +58,9 @@ export class BinaryReader {
    * Reads and validates the header of the KHI file.
    */
   public readHeader(): { magic: string; version: number } {
+    if (this.offset !== 0) {
+      throw new Error('Offset must be 0 to read header');
+    }
     if (this.buffer.byteLength < 4) {
       throw new Error('Buffer too small to contain header');
     }
@@ -93,7 +99,7 @@ export class BinaryReader {
     const compressedData = new Uint8Array(this.buffer, this.offset + 8, size);
 
     // Decompress the gzip data
-    const decompressionStream = new globalThis.DecompressionStream('gzip');
+    const decompressionStream = new DecompressionStream('gzip');
     const sourceBlob = new Blob([compressedData]);
     const textDecompressionStream = sourceBlob
       .stream()
--- a/web/src/app/parser/core/binary-reader.ts
+++ b/web/src/app/parser/core/binary-reader.ts
@@ -1,0 +1,110 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Data associated with a single chunk in the KHI file.
+ */
+export interface ChunkData {
+  readonly size: number;
+  readonly typeId: number;
+  readonly data: Uint8Array;
+}
+
+/**
+ * A utility class to sequentially read binary chunks from a KHI file ArrayBuffer.
+ */
+export class BinaryReader {
+  private readonly dv: DataView;
+  private offset = 0;
+
+  /**
+   * Creates a new BinaryReader for the given buffer.
+   */
+  constructor(private readonly buffer: ArrayBuffer) {
+    this.dv = new DataView(buffer);
+  }
+
+  /**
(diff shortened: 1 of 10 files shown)


web/src/app/parser/core/binary-reader.ts
```diff
@@ -0,0 +1,110 @@
...
+   */
+  public hasMore(): boolean {
+    return this.offset < this.buffer.byteLength;
+  }
+
+  /**
+   * Reads and validates the header of the KHI file.
+   */
+  public readHeader(): { magic: string; version: number } {
+    if (this.buffer.byteLength < 4) {
```
kyasbal, line 64: Maybe it's  better to verify if `this.offset != 0` also?

web/src/app/parser/core/binary-reader.ts
```diff
@@ -0,0 +1,110 @@
...
+    const version = this.dv.getUint8(3);
+    this.offset = 4;
+    return { magic, version };
+  }
+
+  /**
+   * Reads the next chunk metadata and its binary payload, decompressing the payload using gzip.
+   */
+  public async readNextChunk(): Promise<ChunkData> {
+    if (this.offset + 8 > this.buffer.byteLength) {
```
kyasbal, line 86: Let add comment what the 8 means.

web/src/app/parser/core/binary-reader.ts
```diff
@@ -0,0 +1,110 @@
...
+    const typeId = this.dv.getUint32(this.offset + 4, true);
+
+    if (this.offset + 8 + size > this.buffer.byteLength) {
+      throw new Error('Unexpected end of file while reading chunk data');
+    }
+
+    const compressedData = new Uint8Array(this.buffer, this.offset + 8, size);
+
+    // Decompress the gzip data
+    const DecompressionStreamAPI = (
```
kyasbal: It would be better to define `global.d.ts` at the top of web folder? 
Sometimes the default typings for browser APIs are missing if we use modern API. But it shouldn't be used from each part of codes but it should be better to define these additional types at the root.

kyasbal: And it would be better to provide a function like `assertNecessaryAPI` to verify if required new APIs are available on the environment.

web/src/app/parser/core/file-streamer.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Parses the raw KHI binary file into the final UI View Model.
+   */
+  async parse(buffer: ArrayBuffer): Promise<InspectionData> {
+    const reader = new BinaryReader(buffer);
+
+    // 1. Header Validation
+    const { magic, version } = reader.readHeader();
+    if (magic !== 'KHI') {
```
kyasbal: Magic should be defined as constant member at the top of this type.

web/src/app/parser/core/file-streamer.ts
```diff
@@ -0,0 +1,132 @@
...
+    let chunkIndex = 0;
+
+    // 3. Chunk Streaming & Ingestion Phase
+    while (reader.hasMore()) {
+      const offset = reader.currentOffset;
+      const { size, typeId, data } = await reader.readNextChunk();
+
+      const definition = blueprint.get(typeId);
+      if (!definition) {
+        console.warn(
```
kyasbal: I think it's better to throw an error.

web/src/app/parser/core/file-streamer.ts
```diff
@@ -0,0 +1,132 @@
...
+          `Unknown chunk type ${typeId} at offset ${offset}. Skipping.`,
+        );
+        continue;
+      }
+
+      try {
+        // Lazy-load the assembler the first time we encounter this chunk type
```
kyasbal: We don't need to lazy load assembler. It would be better to instantiate all the assemblers for the version at first. Then assembler can throw an error when it had missing chunk for a type.

web/src/app/parser/core/file-streamer.ts
```diff
@@ -0,0 +1,132 @@
...
+        });
+      }
+
+      chunkIndex++;
+    }
+
+    // 4. Priority-Based Assembly Phase
+    // For now, create an empty InspectionData (we need to bypass strict null checks or provide empty arrays/sets)
+    // You may need to adapt this `createEmptyInspectionData` according to your store definition.
+    const finalModel: InspectionData = new InspectionData(
```
kyasbal: I think we define the `InspectionDataBuilder` later as it's written on the README under parser folder?

web/src/app/parser/core/file-streamer.ts
```diff
@@ -0,0 +1,132 @@
...
+    // Sequentially apply data to the final model
+    for (const def of sortedDefinitions) {
+      try {
+        const assembler = activeAssemblers.get(def.typeId)!;
+        assembler.assembleInto(finalModel);
+      } catch (error) {
+        throw new KHIDataAssemblyError({
+          version,
+          typeId: def.typeId,
+          cause: error,
```
kyasbal: These part should be addressed in the InspectionDataBuilder

web/src/app/parser/core/binary-reader.ts
```diff
@@ -0,0 +1,116 @@
...
+    const typeId = this.dv.getUint32(this.offset + 4, true);
+
+    if (this.offset + 8 + size > this.buffer.byteLength) {
+      throw new Error('Unexpected end of file while reading chunk data');
+    }
+
+    const compressedData = new Uint8Array(this.buffer, this.offset + 8, size);
+
+    // Decompress the gzip data
+    assertNecessaryAPI('DecompressionStream');
```
kyasbal: assertNecessaryAPI should be called at the beginning. Otherwise, user will face this error after waiting data loading. It's not userfriendly. It should fail fast.

web/src/app/parser/core/interfaces.ts
```diff
@@ -28,7 +28,7 @@ export interface IDataAssembler<TProto = unknown> {
   /**
    * Integrates the ingested data into the final InspectionData model.
    */
-  assembleInto(builder: InspectionDataBuilder): void;
+  assembleInto(model: InspectionData): void;
```
kyasbal: `InspectionData` is the old data store. We should still rely on the InspectionDataBuilder.

jyane (approved):
LGTM

PR #632, opened 2026-04-28
Add TimelineRegistry and TimelinePath

kyasbal:
This PR adds `TimelinePath`, `TimelinePathPool`, `TimelineRegistry`, `TimelineBuilder`.

* The `TimelinePath` is a linked list data structure representing Timeline location like `Pod(kind) -> default(namespace) -> nginx(resource)`. Each timelines are save with associating them, thus this needs to be singleton in `TimlinePathPool`.
* The `TimelineRegistry` holds `TimelineBuilder` associating them with the `TimelinePath`. Log parsers will get a builder from `TimelineRegistry` with specifying a `TimelinePath` to modify timeline contents.

### DIFF
--- a/pkg/model/khifile/v6/timeline_path.go
+++ b/pkg/model/khifile/v6/timeline_path.go
@@ -1,0 +1,148 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package khifilev6
+
+import (
+	"iter"
+	"sync"
+
+	pb "github.com/GoogleCloudPlatform/khi/pkg/generated/khifile/v6"
+	"google.golang.org/protobuf/proto"
+)
+
+// TimelinePath represents a single node in the timeline hierarchy tree.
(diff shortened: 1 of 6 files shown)


pkg/model/khifile/v6/timeline_path.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	return current
+}
+
+// getOrCreateSingle handles the thread-safe retrieval or creation of a single TimelinePath segment.
+func (p *TimelinePathPool) getOrCreateSingle(parent *TimelinePath, name string, t *pb.TimelineType) *TimelinePath {
+	nameRef := p.stringPool.InternString(name)
+
+	if t == nil || t.Id == nil {
+		panic("TimelineType and its ID must not be nil")
```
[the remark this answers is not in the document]
kyasbal: This is allowed because it only happens when the codebase contains the TimelineType not corectly imported. But it's better to show more details on the message.
User may need to register the TimelineType ID correctly to assign the right ID.

pkg/model/khifile/v6/timeline_path.go
```diff
@@ -0,0 +1,148 @@
...
+	return current
+}
+
+// getOrCreateSingle handles the thread-safe retrieval or creation of a single TimelinePath segment.
+func (p *TimelinePathPool) getOrCreateSingle(parent *TimelinePath, name string, t *pb.TimelineType) *TimelinePath {
+	nameRef := p.stringPool.InternString(name)
+
+	if t == nil || t.Id == nil {
+		panic("TimelineType and its ID must not be nil. TimelineType must be correctly registered to assign the right ID.")
+	}
```
RyuSA, line 93: Can this if statement be moved to the beginning?

RyuSA (approved):
LGTM, I left a small comment.

pkg/model/khifile/v6/timeline_path.go
kyasbal: Done/

PR #633, opened 2026-04-28
Add LogAccumulator storing all logs before serializing them into the khifile

kyasbal:
This change adding `LogAccumulator`.

This type receives logs that will be serialized into the KHI file from parsers and returns the list when the file generator needed it.

### DIFF
--- a/pkg/model/khifile/v6/ids.go
+++ b/pkg/model/khifile/v6/ids.go
@@ -28,6 +28,12 @@ const (
 	IDString IDNamespace = iota
 	// IDFieldSet is the namespace for field set IDs.
 	IDFieldSet
+	// IDTimelinePath is the namespace for timeline path IDs.
+	IDTimelinePath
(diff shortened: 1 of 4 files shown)


pkg/model/khifile/v6/ids.go
```diff
@@ -28,6 +28,12 @@ const (
 	IDString IDNamespace = iota
 	// IDFieldSet is the namespace for field set IDs.
 	IDFieldSet
+	// IDTimelinePath is the namespace for timeline path IDs.
+	IDTimelinePath
+	// IDTimelineItems is the namespace for timeline items IDs.
+	IDTimelineItems
```
RyuSA, line 34: note: this part will be conflicted. #632

RyuSA (approved):
LGTM

pkg/model/khifile/v6/ids.go
kyasbal: I'll resolve it manually. That's gonna be fine.

PR #640, opened 2026-05-01
Added interned pool types and struct decoder

kyasbal:
This PR the domain layer data store for the interned values.

These domain types receives values converted from proto types, store them in memory efficient method because store needs to keep them in memory even user is not currently seeing the data on the screen, and provides efficient methods to access data.

`InternedString` and `InternedStructDecoder` are one of the domain type but not expected to be directly touched from the view. These are used from TimelineStore or LogStore that will be implemented later.


web/src/app/store/domain/intern-pool-store.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+export interface IRawFieldPathSetEntry {
+  /**
+   * The unique ID of the field path set.
+   */
+  readonly id: number;
+  /**
+   * The array of string IDs representing the segments of field paths.
+   */
+  readonly fieldNames: readonly number[];
```
jyane: This seems not a string array? If this is intended (as per tests?), could we add more detailed explanation here?

web/src/app/store/domain/intern-pool-store.ts
```diff
@@ -0,0 +1,108 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Represents an entry in the interned string pool.
+ */
+export interface IRawStringEntry {
```
jyane: Does this `I` stand for interface? I feel this is a bit weird to see here in typescript.

web/src/app/store/domain/intern-pool-store.ts
kyasbal: it's array of string IDs. Although it's written on the comment, indeed the  field name is misleading. I'll patch it.

web/src/app/store/domain/intern-pool-store.ts
kyasbal: You're right. I hadn't really thought about it, but it doesn't match the other code, so I'll fix it.

jyane (approved):

PR #641, opened 2026-05-01
Adding log store domain layer

kyasbal:
This PR adds log store in the domain layer. 

These domain types receives values converted from proto types, store them in memory efficient method because store needs to keep them in memory even user is not currently seeing the data on the screen, and provides efficient methods to access data.

Any views will access to Log data or the Style data from these domain layer.


kyasbal:
I updated this PR regarding the file schema spec #650.

kyasbal:
I'll split this into 2 different PRs.

renamoo (approved):

PR #680, opened 2026-05-21
Added timeline store type

kyasbal:
Adding domain type for TimelineStore.

This is the central domain type holding all the timeline structure.


kyasbal:
See #681 to know how the type is used.

PR #692, opened 2026-05-23
fix: confirm refresh when backend disconnects

### LINKED ISSUE #179: Show popup to confirm window refresh if the backend server stopped working
### Is your feature request related to a problem? Please describe

If the backend server stopped or died for some reason, the data of the client side can be lost when the user refreshes the browser window.

### Describe the solution you'd like

Once the client is not able to reach the backend server or the backend server returns some specific error response, we want the client to show the browser-default popup which asks a user if they want to proceed with closing or refreshing the window. This is possible by capturing [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) event and calling `preventDefault()`.

Haihan-Jiang:
## Summary
- Listen to the existing backend connection status in the root component.
- Register the browser native `beforeunload` confirmation only while the backend is disconnected.
- Remove the handler when the backend reconnects or the root component is destroyed.

Fixes #179

## Tests
- `npx ng run frontend:test --browsers=ChromeHeadlessNoSandbox --watch=false --include=src/app/root.component.spec.ts`
- `npm run lint`
- `npx ng run frontend:build:dev`
- `git diff --check`

### DIFF
--- a/web/src/app/root.component.spec.ts
+++ b/web/src/app/root.component.spec.ts
@@ -72,20 +72,21 @@ describe('RootComponent', () => {
     expect(event.returnValue).toBe('');
   });
 
-  it('should remove confirmation before unload when backend reconnects', () => {
+  it('should skip confirmation before unload when backend is connected', () => {
     const fixture = TestBed.createComponent(RootComponent);
-    fixture.detectChanges();
-    connectionStatus.set(BackendConnectionStatus.Disconnected);
     fixture.detectChanges();
     const handler = getBeforeUnloadHandler(addEventListenerSpy);
 
     connectionStatus.set(BackendConnectionStatus.Connected);
     fixture.detectChanges();
+    const event = {
+      preventDefault: jasmine.createSpy('preventDefault'),
+      returnValue: 'initial',
+    } as unknown as BeforeUnloadEvent;
+    handler(event);
 
-    expect(removeEventListenerSpy).toHaveBeenCalledWith(
-      'beforeunload',
-      handler,
-    );
+    expect(event.preventDefault).not.toHaveBeenCalled();
+    expect(event.returnValue).toBe('initial');
   });
 
   it('should remove confirmation before unload when destroyed', () => {
--- a/web/src/app/root.component.spec.ts
+++ b/web/src/app/root.component.spec.ts
@@ -1,0 +1,115 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


kyasbal (changes requested):
@Haihan-Jiang  Hello. Thank you for your contribution.
Your code is clean enough and align with our coding rules.

However, the functionality implemented in this PR doesn't make sense.
Your logic is basically enabling/disabling the beforeunload event handler in the response to the connection status change. 

As far as I remember, the initial motivation for the issue #179 is for preventing losing loaded data on client if the server is dead by OOM or other errors and user reloaded it without noticing it.
So maybe what we should do is:
* Show a dialog confirming the close only when the server is disconnected. 
* If the server is connected healthy, then the client shouldn't confirm the closing. Because user should be able to see the same inspection data again with reopening the URL.

Haihan-Jiang:
Thanks for the clarification. I updated this to match that behavior more directly in 8407f45: the beforeunload listener is registered once, and the handler checks the current backend connection status when the unload event fires. It only calls preventDefault/sets returnValue when the status is BackendConnectionStatus.Disconnected; healthy/connected states return without confirmation.

I also updated the root component specs for disconnected, connected, and destroy behavior.

Local validation:
- npm run lint -- --fix=false
- npx ng run frontend:test --watch=false --browsers=ChromeHeadless --include=src/app/root.component.spec.ts

Haihan-Jiang:
Hi @kyasbal, I updated the implementation in 8407f45 to register the beforeunload listener once and only prompt when the latest backend status is disconnected. The targeted root component tests were updated as well, and the current PR checks are green except for the project-side ondemand deploy check. Could you take another look when you have a chance?

kyasbal:
I'll make time to check it on Monday. Sorry for the late reply.

Haihan-Jiang:
No worries, thanks for the update. Monday works for me.

kyasbal (approved):
I tested this on my environment and it works as expected.
Thank you for your first contribution.

kyasbal:
@Haihan-Jiang  I can't merge this now because your commit is not signed. 

References:
https://github.com/GoogleCloudPlatform/khi/blob/main/docs/en/development-contribution/development-guide.md#setup-commit-signature-verification
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Can you configure your git client to sign commits and push it again? The code itself looks good thus we'll merge this once your commit has verified signs. This also applies for the other PRs opened by you.

Haihan-Jiang:
Thanks, I see the blocker. The code approval is clear; I am setting up verified commit signing and will repush the branch once the signing key is active.

kyasbal:
Now I approved merging this PR. It will be automatically merged soon.
Thank you for your first contribution on our project😃

PR #709, opened 2026-05-30
Document KHI job mode

### LINKED ISSUE #196: Document KHI Job Mode
### Description

The KHI tool has a "job mode" that allows it to be invoked directly from the command-line interface (CLI) to generate KHI files. This functionality is distinct from launching KHI via the Web UI and is particularly useful for integrations, such as those with alerting systems.

While this feature has been implemented, there is currently no existing documentation for it. This issue proposes the creation of comprehensive documentation for the KHI job mode, detailing its usage, parameters, and any relevant examples.

---

### Action Items

* Document the command-line interface for KHI job mode.
* Provide examples of how to generate KHI files using the CLI.

Haihan-Jiang:
## Summary
- add a setup guide for running KHI in job mode
- document the required job mode flags, supported inspection type IDs, and a basic GKE example
- link the new guide from the README user guide section

Fixes #196

## Tests
- `go test ./pkg/parameters`
- `git diff --check`

### DIFF
--- a/docs/en/setup-guide/job-mode.md
+++ b/docs/en/setup-guide/job-mode.md
@@ -32,7 +32,7 @@ Available inspection type IDs include:
   --job-mode \
   --job-inspection-type gcp-gke \
   --job-inspection-features <feature-id>,<feature-id> \
-  --job-inspection-values '{"<form-field-id>":"<value>"}' \
+  --job-inspection-values '{"<text-field-id>":"<value>","<set-field-id>":["<value>"]}' \
   --job-export-destination ./inspection.khi
 ```
 
@@ -58,7 +58,7 @@ For example, a Google Cloud inspection usually includes common fields such as pr
 }
 ```
 
-The exact set of required fields depends on the inspection type and enabled features. The most reliable way to build the JSON is to run the same inspection once in the web UI, open the browser developer tools, and check the request payload sent to `POST /api/v3/inspection/<inspection-id>/run` or `POST /api/v3/inspection/<inspection-id>/dryrun`. The `parameters` object in that request uses the same field IDs and values that job mode expects.
+The exact set of required fields depends on the inspection type and enabled features. The most reliable way to build the JSON is to run the same inspection once in the web UI, open the browser developer tools, and check the request payload sent to `POST /api/v3/inspection/<inspection-id>/run` or `POST /api/v3/inspection/<inspection-id>/dryrun`. Copy the `parameters` object from that request as the `--job-inspection-values` JSON. It uses the same field IDs and values that job mode expects.
 
 For a GKE inspection, the commonly required field IDs are:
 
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review from my side. The remaining github-deploy-ondemand check appears to require project-side approval/run permissions; could someone take a look or trigger it when convenient?

docs/en/setup-guide/job-mode.md
```diff
@@ -0,0 +1,104 @@
...
+## **Example**
+
+The following example runs a GKE inspection and writes the result to `./gke-inspection.khi`:
+
+```bash
+./khi \
+  --job-mode \
+  --job-inspection-type gcp-gke \
+  --job-inspection-features ALL \
+  --job-inspection-values '{
```
kyasbal, line 141: I think the most difficult part of this feature is building the `--job-inspection-values` parameter.
The current easiest way is using dev tool on browser to see what is sent from the frontend. It's a little hacky way but worth to be documented.

docs/en/setup-guide/job-mode.md
```diff
@@ -0,0 +1,104 @@
...
+    "cloud.google.com/common/input-project-id": "my-project",
+    "cloud.google.com/common/input-location": "us-central1",
+    "cloud.google.com/common/input-end-time": "2026-01-15T10:00:00Z",
+    "cloud.google.com/common/input-duration": "2h",
+    "<inspection-specific-field-id>": "<value>"
```
kyasbal: It would be better if we can show the example workable job command with the full parameter.
The current example lacks of many parameters and user need to find what are missing. It'll be nicer if we can list keys required for actually running the job mode. Values for the parameters can be placeholder then.

kyasbal (changes requested):

Haihan-Jiang:
Thanks, I updated the job mode doc in two places:

- added a practical way to build `--job-inspection-values` from the browser dev tools request payload for `/run` or `/dryrun`
- expanded the GKE example to include concrete common field IDs such as project, location, cluster name, namespaces, and kinds

I kept the wording scoped because the exact required fields still depend on the selected inspection type and enabled features.

kyasbal:
The current example parameter list is still missing several fields. How about asking your codex to run the job to test?

Haihan-Jiang:
Thanks, I asked Codex to run the job-mode command and it found two concrete issues in the example.

Fixed in `18e401e`:

- set-form values such as `cloud.google.com/k8s/input-namespaces` and `cloud.google.com/k8s/input-kinds` must be JSON arrays, not scalar strings. The previous example fails with `request parameter cloud.google.com/k8s/input-kinds#default was not given in array`.
- `ALL` enables optional GKE features that are disabled by default in the UI, so the runnable default GKE example now uses an explicit feature ID list. I also added the extra fields needed when optional features such as node, container, control plane, CSM, and serial port logs are selected.

Validation performed locally:

- `npx markdownlint-cli2 docs/en/setup-guide/job-mode.md` -> 0 errors.
- Re-ran job mode with the corrected payload and explicit default GKE feature list. It now gets past the form parameter parsing and stops at local ADC setup: `failed to create monitoring metric client: credentials: could not find default credentials`. I do not have ADC configured in this local environment, so I could not complete a Cloud Logging-backed export here, but the prior parameter-shape failure is fixed.

docs/en/setup-guide/job-mode.md
Haihan-Jiang: Addressed in `18e401e`. The guide now keeps the browser dev tools `/run` or `/dryrun` payload workflow, clarifies that set-form fields must be JSON arrays, and updates the GKE example to use the same field IDs/value shapes that job mode expects. I also added a table of extra fields required when optional GKE features are enabled.

docs/en/setup-guide/job-mode.md
Haihan-Jiang: Also addressed in `18e401e`. I replaced the `ALL` example with an explicit default GKE feature list and a payload that uses the full required field set for that selected feature set. The guide now also lists the extra field keys to include when optional GKE features are enabled or when `ALL` is used.

kyasbal:
I'm sorry to return this back again and again. However, this example query must generate the empty result even if the target project and cluster actually exist because resource names input are missing.

We can find the right parameters from chrome inspector when we fills parameters on the form.
image

Thus parameters beginning with `cloud.google.com/common/input-query-resource-names` would be necessary to be filled with `projects/<the-proejct-id>`. This is not great from user experience perspective but for considering the project forwarding its logs to the other projects, KHI receives target project for each queries. These are usually optional field on the UI and automatically filed from the other project ID field. But it's still required for the job mode. Indeed this behavior must be improved in near future...

kyasbal:
We had added the feature to show job command on the new inspection dialog. So the situation around job mode was changed recently. I'm sorry to close this PR but thank you for trying to contribute.

PR #750, opened 2026-06-09
Added sorting configuration on timeline type

kyasbal:
This PR adds additional sorting settings in TimelineTypes. It allows custom sorting among multiple timelines with the same sort priority.


kyasbal:
Removed few unit tests here because these will be replaced in next another PR.

PR #798, opened 2026-06-22
ci: enforce prettier formatting for .cloudbuild

### LINKED ISSUE #791: Enforce formatter under .cloudbuild
Add formatter for .cloudbuild/. Otherwise it always produce diffs unintentionally.

manumishra12:
## Summary

Fixes #791. `.cloudbuild/*.yaml` had no formatter, so edits to those files produced unintentional diffs. This adds prettier-based formatting + enforcement, mirroring the existing `format-web` / `check-format-web` pattern.

## Changes

- **`scripts/make/lint.mk`**: add `format-cloudbuild` (`prettier --write`) and `check-format-cloudbuild` (`prettier --check`) targets for `.cloudbuild/**/*.yaml`.
- **`Makefile`**: wire `format-cloudbuild` into the aggregate `format` target.
- **`.github/workflows/pullrequest.yaml`**: run `make check-format-cloudbuild` in the `backend-tests` job (it already runs `npm ci` + `check-format-go`).
- **`package.json` / `package-lock.json`**: add `prettier` as a root devDependency. The root package only had `markdownlint-cli2`; the web package's prettier isn't available to root-level make targets, so the root needs its own (pinned to match web, `^3.8.1`).

## Note

The existing `.cloudbuild` YAML files already conform to prettier, so there's **no reformatting** in this PR — it only adds the tooling/enforcement so future edits stay consistent.

## Verification

```
make check-format-cloudbuild   # passes on current files
make format-cloudbuild         # no-op (already formatted)
```


kyasbal:
Can I ask you to include other setting files that is not target of any other formatters in the prettier target as well?
There are several setting files under `.github` or `.vscode` as well. I think it would be better to have `make format-misc` or `make format-check-misk` to consolidate all the formatting task for the other configuration files.

I only mentioned about .cloudbuild in the original issue and this is outside of what I wrote in the issue. The change itself is okey, so if you want, I can merge this as it is.

manumishra12:
Thanks @kyasbal — good call. I generalized the target to `format-misc` / `check-format-misc`, covering `.cloudbuild`, `.github` and `.vscode` (yaml/yml/json), wired into `make format` and the `backend-tests` CI job.

Heads-up: enabling prettier on `.github` reformatted the workflow files (mostly quote style `'` → `"` and YAML sequence indentation — prettier defaults, matching the existing web formatter; no semantic changes). `.vscode/settings.json` was reindented too. Happy to tweak (e.g. add a `.prettierrc`) if you'd prefer different options.

package.json
```diff
@@ -6,6 +6,7 @@
   "copyright": "Copyright 2025 Google LLC",
   "private": true,
   "devDependencies": {
-    "markdownlint-cli2": "^0.19.0"
+    "markdownlint-cli2": "^0.19.0",
+    "prettier": "^3.8.1"
```
kyasbal: We have `3.8.4` in package.lock.json. Then this should be `^3.8.4`, shouldn't it?
Or do you have purpose to use `^3.8.1` instead of `^3.8.4`?

kyasbal (changes requested):

manumishra12:
Good catch, @kyasbal — no, there was no particular reason for `^3.8.1`. You're right that the lockfile resolves prettier to `3.8.4`, so I've bumped the range to `^3.8.4` in both `package.json` and the lockfile's `devDependencies` entry to match. Pushed.

kyasbal (approved):
LGTM

kyasbal:
Thank you for working on this PR! Merging.

PR #826, opened 2026-07-03
Change COOP header from same-origin to same-origin-allow-popups

kyasbal:
This allows plugins to open a popup from the other site like the popup to use Google auth libraries.


kyasbal:
This will break SharedArrayBuffer usage on the frontend. I'll reconsider the other way.

PR #829, opened 2026-07-06
feat(job-mode): support file input tasks in job mode CLI command generation

### LINKED ISSUE #827: feat(job-mode): support file input tasks in job mode CLI command generation
## Background & Problem
PR #821 introduced the `jobCommand` metadata and front-end component to display a copy-pasteable CLI command representation of the configured query parameters for KHI Job mode.

However, inspection types that accept file inputs (such as OSS Kubernetes audit log inspection via `InputAuditLogFilesTask` / `FileFormTask`) are not currently supported in Job mode command generation. When using file-based inspection types, the generated Job mode command fails to represent local file inputs.

## Desired Behavior
- Job mode CLI execution should support specifying local log file paths for file-input inspection tasks (e.g., OSS Kubernetes audit logs).
- The generated CLI command example in the UI (`khi-job-command`) should correctly include arguments or parameter representations for local file paths (e.g., `--job-inspection-values` containing file paths or placeholder path flags).

## Proposed Tasks
1. Extend `JobModeCommandTask` / `GenerateJobModeCommand` to handle file-based input tasks (`FileFormTask`).
2. Update Job mode argument parsing in the backend runner to accept local file path values for `FileFormTask` parameters.
3. Verify and update the frontend UI (`khi-job-command`) to render valid CLI command examples for file-input inspection types.

Ackberry:
Adds support for file-input inspection types (like OSS Kubernetes audit logs) in Job mode.

Before this change, the Job-mode command generated in the New Inspection dialog couldn't represent local file inputs, so file-based inspections couldn't be reproduced headlessly. This PR makes the generated command include a placeholder path for file fields, and makes the Job-mode runner load a local file from that path.

The main pieces:
- `GenerateJobModeCommand` now emits a placeholder path in `--job-inspection-values` for each file field, and a new `FormFieldSetMetadata.GetFileFieldIDs()` lets it find which fields are files.
- A new `LocalFileUploadToken` and `InPlaceUploadFileStoreProvider` read the file at its given path in place.
- `UploadFileStore` now picks a provider based on the token's `GetType()`, so local-file tokens read through the in-place provider while the existing browser-upload flow is unchanged. The store is now set up in both server and job mode.
- `FileFormTask` loads a local path from the request values when running in Job mode, and otherwise falls back to the existing upload flow.

Testing: added unit tests for the in-place provider read, the command generation with a file field, and the local-file load-and-verify path. The full suite passes with `go test ./... -args -skip-cloud-logging=true`, and I verified end to end by running an OSS file-input inspection in Job mode against a local JSONL file and confirming it produced a .khi output.

Frontend work is not done in this PR. The command preview already renders the placeholder correctly since the component just displays whatever the backend generates, so no frontend change is needed for correctness. An optional hint telling users to replace the placeholder path could be a follow-up.

closes #827


kyasbal:
@Ackberry Thank you for working on this.

image

Could you configure your git client to sign on your commits? We require every commit to be signed.
https://docs.github.com/en/authentication/managing-commit-signature-verification

pkg/server/upload/store.go
```diff
@@ -121,23 +122,23 @@ func (s *UploadFileStore) SetResultOnCompletedUpload(token UploadToken, uploadEr
...
+			StoreProvider:     s.providerForToken(token),
 			Status:            UploadStatusWaiting,
 			UploadError:       uploadError,
 			VerificationCount: prev.VerificationCount,
 		}
 	}
 	if uploadError == nil {
 		go func() {
-			err := s.verifiers[token.GetID()].Verify(s.StoreProvider, token)
+			err := s.verifiers[token.GetID()].Verify(s.providerForToken(token), token)
```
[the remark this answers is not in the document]
kyasbal: This suggestion looks valid. Could you address this or explain why you think it doesn't need to address?

pkg/core/inspection/formtask/fileform.go
```diff
@@ -57,6 +57,12 @@ func (b *FileFormTaskBuilder) Build(labelOpts ...common_task.LabelOpt) common_ta
 	return common_task.NewTask(b.FormTaskBuilderBase.id, b.FormTaskBuilderBase.dependencies, func(ctx context.Context) (upload.UploadResult, error) {
 		metadata := khictx.MustGetValue(ctx, inspectioncore_contract.InspectionRunMetadata)
 
+		req := khictx.MustGetValue(ctx, inspectioncore_contract.InspectionTaskInput)
+		if pathValue, ok := req[b.FormTaskBuilderBase.id.ReferenceIDString()].(string); ok && pathValue != "" {
+			token := &upload.LocalFileUploadToken{FilePath: pathValue}
+			return upload.DefaultUploadFileStore.LoadLocalFileResult(token, b.verifier), nil
+		}
```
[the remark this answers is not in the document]
kyasbal: This is a security problem.

If we just passes string file path in the task input, then KHI will just read the file even if it was running as a web server. It will allow reading any files on the file system.

Also we should consider a way to generalize these part? It's a not great idea to depend parameters.Job.JobMode from here. But we can have the StoreProvider only registered in the job mode.

cmd/kubernetes-history-inspector/main.go
```diff
@@ -147,6 +147,16 @@ func run() int {
...
 
+	uploadFileStoreFolder := "/tmp"
+
+	if parameters.Common.UploadFileStoreFolder != nil {
+		uploadFileStoreFolder = *parameters.Common.UploadFileStoreFolder
+	}
+
+	upload.DefaultUploadFileStore = upload.NewUploadFileStore(upload.NewLocalUploadFileStoreProvider(uploadFileStoreFolder))
+
+	upload.DefaultUploadFileStore.RegisterProvider((&upload.LocalFileUploadToken{}).GetType(), &upload.InPlaceUploadFileStoreProvider{})
```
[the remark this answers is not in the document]
kyasbal: This also looks valid and this is very linked to another comment:

https://github.com/GoogleCloudPlatform/khi/pull/829/changes#r3527279385

kyasbal (changes requested):

cmd/kubernetes-history-inspector/main.go
```diff
@@ -147,6 +147,16 @@ func run() int {
...
 	go handleTerminateSignal(exitCh)
 
+	uploadFileStoreFolder := "/tmp"
+
+	if parameters.Common.UploadFileStoreFolder != nil {
+		uploadFileStoreFolder = *parameters.Common.UploadFileStoreFolder
+	}
+
+	upload.DefaultUploadFileStore = upload.NewUploadFileStore(upload.NewLocalUploadFileStoreProvider(uploadFileStoreFolder))
+	if *parameters.Job.JobMode {
```
kyasbal: `if !*parameters.Job.JobMode {} else {}` clause follows. Why don't you merge this logic in the following else clause?

pkg/task/inspection/inspectioncore/impl/job_command_task.go
```diff
@@ -67,16 +73,28 @@ var JobModeCommandTask = inspectiontaskbase.NewInspectionTask(
...
 // GenerateJobModeCommand formats a copy-pasteable command to execute KHI in job mode.
-func GenerateJobModeCommand(inspectionType string, enabledFeatures []string, taskInput any) (string, error) {
+func GenerateJobModeCommand(inspectionType string, enabledFeatures []string, taskInput any, fileFieldIDs []string) (string, error) {
 	// Create a copy to avoid mutating the original features array.
 	featuresCopy := make([]string, len(enabledFeatures))
 	copy(featuresCopy, enabledFeatures)
 	slices.Sort(featuresCopy)
 	featuresStr := strings.Join(featuresCopy, ",")
 
+	values, _ := taskInput.(map[string]any)
```
kyasbal: With addressing the PR comment(https://github.com/GoogleCloudPlatform/khi/pull/829/changes#r3532721801), this cast will be no longer needed.

pkg/task/inspection/inspectioncore/impl/job_command_task.go
```diff
@@ -67,16 +73,28 @@ var JobModeCommandTask = inspectiontaskbase.NewInspectionTask(
...
+	if len(fileFieldIDs) > 0 {
+		merged := make(map[string]any, len(values)+len(fileFieldIDs))
+		for k, v := range values {
+			merged[k] = v
+		}
+		for _, id := range fileFieldIDs {
+			merged[id] = "path/to/file"
+		}
+		values = merged
+	}
```
kyasbal, line 90: This would be better to be simplified.
```suggestion
    merged := maps.Clone(taskInput)
    for _, id := range fileFieldIDs {
       merged[id] = "path/to/file"
    }
```

pkg/task/inspection/inspectioncore/impl/job_command_task.go
```diff
@@ -55,7 +55,13 @@ var JobModeCommandTask = inspectiontaskbase.NewInspectionTask(
 			taskInput = input
 		}
```
kyasbal: This was coming from my previous commit, but here can be simplified. Then the `taskInput` become `map[string]any` not `any`. It will require some changes in `GenerateJobModeCommand` function argument type.

```suggestion
taskInput := khictx.MustGetValue(ctx, inspectioncore_contract.InspectionTaskInput)
```

pkg/core/inspection/formtask/fileform.go
```diff
@@ -57,6 +58,13 @@ func (b *FileFormTaskBuilder) Build(labelOpts ...common_task.LabelOpt) common_ta
...
 
+		req := khictx.MustGetValue(ctx, inspectioncore_contract.InspectionTaskInput)
+		if *parameters.Job.JobMode {
+			if pathValue, ok := req[b.FormTaskBuilderBase.id.ReferenceIDString()].(string); ok && pathValue != "" {
+				token := &upload.LocalFileUploadToken{FilePath: pathValue}
+				return upload.DefaultUploadFileStore.LoadLocalFileResult(token, b.verifier), nil
+			}
+		}
 		token := upload.DefaultUploadFileStore.GetUploadToken(GenerateUploadIDWithTaskContext(ctx, b.FormTaskBuilderBase.id.ReferenceIDString()), b.verifier)
 		uploadResult, err := upload.DefaultUploadFileStore.GetResult(token)
```
kyasbal: How about changing the signature of GetResult to accept the `req` as well?
Then the store can read the file contents from request parameter.
In the case, we don't need to depend on parameter in this task code, and we can just consolidate all codes depending on parameter.Job.JobMode in `main.go` with injecting the custom data store implementation for job mode.

kyasbal (changes requested):

Ackberry:
GetResult now takes the request values, and `main.go` injects a JobModeUploadFileStore in job mode that resolves the local file path from them. `fileform.go` no longer depends on parameters.Job.JobMode, and LoadLocalFileResult is removed.

cmd/kubernetes-history-inspector/main.go
```diff
@@ -200,6 +200,7 @@ func run() int {
 
 		displayStartMessage(*parameters.Server.Host, *parameters.Server.Port)
 	} else {
+		upload.DefaultUploadFileStore = upload.NewJobModeUploadFileStore(uploadFileStore)
```
kyasbal: nit: This should be done after the logging at the following line

pkg/server/upload/jobmodestore.go
```diff
@@ -0,0 +1,65 @@
...
+// inputs from local file paths in the request values when running in job mode.
+func (s *JobModeUploadFileStore) GetResult(token UploadToken, req map[string]any) (UploadResult, error) {
+	s.fieldIDLock.RLock()
+	fieldID := s.fieldIDs[token.GetID()]
+	s.fieldIDLock.RUnlock()
+
+	pathValue, ok := req[fieldID].(string)
+	// No local path in the request values: fall back to the normal upload flow.
+	if !ok || pathValue == "" {
+		return s.UploadFileStore.GetResult(token, req)
```
kyasbal: Falling back to UploadFileStore may practically not be usable?
The JobMode won't run the web server and user can't upload file on it.

It should return an error from Job mode side. Otherwise, the upload file store just return some unrelated error.

pkg/server/upload/jobmodestore.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	localToken := &LocalFileUploadToken{FilePath: pathValue}
+
+	s.verifierLock.RLock()
+	verifier, found := s.verifiers[token.GetID()]
+	s.verifierLock.RUnlock()
+	if !found {
+		return s.UploadFileStore.GetResult(token, req)
```
kyasbal: Same.

Falling back to UploadFileStore may practically not be usable?
The JobMode won't run the web server and user can't upload file on it.

It should return an error from Job mode side. Otherwise, the upload file store just return some unrelated error.

pkg/server/upload/jobmodestore.go
```diff
@@ -0,0 +1,65 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package upload
+
+// JobModeUploadFileStore is an UploadFileStore variant for job mode.
+// It resolves file inputs from local file paths in the inspection request
+// values instead of waiting for browser uploads.
+type JobModeUploadFileStore struct {
+	*UploadFileStore
```
kyasbal: Why this can't be just JobModeStore?
JobMode won't run the web server and it can't receive the uploaded file.

kyasbal (changes requested):

kyasbal:
@Ackberry I read the code again. The approach to use the store looks better than the code before modifying the file form task. But the JobModeUploadFileStore seems to have unnecessary dependency on UploadFileStore.
Job mode won't run the web server thus, the upload part won't function properly and we don't need the upload feature on the job mode.

kyasbal:
I was a bit busy and couldn't check it, but I'll make time to check it on Monday. It seems pretty good overall.

Ackberry:
no worries! have a great weekend😄

kyasbal (approved):
Finally it's LGTM 🎉 Thank you for working on this PR!

kyasbal:
I tested this build with my environment as following and it works.

```shell
$ docker run -p 8080:8080 -v /home/ikakeru/jsonl:/jsonl gcr.io/kubernetes-history-inspector/develop:88d2ad761a0bdeae06195ad13a87a04a8001bf7f    --job-mode   --job-inspection-type="oss-kubernetes-from-files"   --job-inspection-features="khi.google.com/k8s-common-auditlog/k8s-auditlog-parser-tail#oss,khi.google.com/oss/event-timeline-mapper#default"   --job-inspection-values='{
  "khi.google.com/oss/form/kube-apiserver-audit-log-files": "/jsonl/audit-20250613181117.jsonl",
  "timezoneShift": 9
}'   --job-export-destination="/jsonl/output.khi"
```

PR #874, opened 2026-08-06
fix: remove CSM traffic log from the container log filter

K53:
Fixed the bug that caused CSM Traffic logs to be duplicated in k8s container logs.

Fixed:
image


kyasbal (approved):
LGTM

PR #928, opened 2026-08-31
refactor(structured,parsers): optimize log parsing with FieldPath interning and Extractor pattern

kyasbal:
## Summary

This PR introduces typed log extraction optimizations and structured log parsing enhancements across all inspection log parsers in KHI:

1. **Pre-compiled `FieldPath` & Interning**:
   - Introduces `structured.CompileFieldPath` with `unique.Handle` string interning to avoid repeated allocations and path parsing during hot log inspection loops.
   - Migrates log readers and models to use package-level `CompileFieldPath` definitions.

2. **Extractor Pattern & Typed Parsing**:
   - Migrates K8s audit log parsers, GCP audit log parsers, and remaining log parsers (GKE Autoscaler, serial port, K8s events/control-plane/node/containers, etc.) to the typed Extractor pattern.
   - Replaces untyped map-based log parsing with high-performance typed access via `NodeReader`.

3. **Testing Infrastructure**:
   - Introduces `structured.MockNode` and test utilities for fast, deterministic unit testing of log parsers without JSON serialization overhead.
   - Migrates unit tests across log parsers to use table-driven tests with `MockNode` and `cmp.Diff`.

## Verification
- `make pre-commit` passed cleanly.
- `make lint-web` passed cleanly.
- `make test-web` passed cleanly (1066/1066 tests).
- `make test-go` passed cleanly across all backend packages.
- `make build-go` and `make build-web` passed cleanly.


.agents/skills/khi_parser/SKILL.md
```diff
@@ -115,27 +109,29 @@ var InputFilterKeywordTaskID = taskid.NewDefaultImplementationID[string](TaskIDP
...
 
 ```go
 package customapp_contract
 
 import (
  "github.com/GoogleCloudPlatform/khi/pkg/common/structured"
- "github.com/GoogleCloudPlatform/khi/pkg/model/log"
+)
+
+var (
```
kyasbal, line 141: Please note that these path is expected to be constant and placed as a package variable. It should be placed just below of the imports.

.agents/skills/khi_parser/SKILL.md
```diff
@@ -145,29 +141,14 @@ type CustomAppFieldSet struct {
...
-
-// Read parses fields from structured log data.
-func (r *CustomAppFieldSetReader) Read(reader *structured.NodeReader) (log.FieldSet, error) {
- return &CustomAppFieldSet{
-  AppName:   reader.ReadStringOrDefault("app_name", "unknown-app"),
-  RequestID: reader.ReadStringOrDefault("request_id", ""),
-  Payload:   reader.ReadStringOrDefault("payload", ""),
+// ExtractCustomApp extracts CustomAppFieldSet from a raw log node reader.
+func ExtractCustomApp(reader *structured.NodeReader) (CustomAppFieldSet, error) {
+ return CustomAppFieldSet{
```
kyasbal, line 159: ExtractXXX is expected to return non-pointer value. If every extract methods are using pointer values, these would be called massively multiple counts and pressure heap memory

.agents/skills/khi_parser/SKILL.md
```diff
@@ -416,22 +387,13 @@ func (i *CustomAppLogIngester) ProcessLog(ctx context.Context, l *log.Log) (*khi
...
-  // Log related parsing logic. LogType, Timestamp, Severity and Summary must be set.
 
- cs.SetLogType(foocontract.LogTypeFoo)
- // Explicitly retrieve and ingest Timestamp from CommonFieldSet.
- if commonFS, err := log.GetFieldSet(l, &log.CommonFieldSet{}); err == nil {
-  cs.SetTimestamp(commonFS.Timestamp)
-  // Do not use commonFS.Severity because it's deprecated.
- }
+ cs.SetLogType(customapp_contract.LogTypeCustomApp)
+ // Timestamp is already populated on l.Timestamp during ingestion.
```
kyasbal: It's usually a timestamp of the log. But not necessary. 
For example, if the log contains its own custom field and it contains precise timestamp rather than the log default timestamp. It would have benefit to set its own value.

.agents/skills/log_timeline_mapper/SKILL.md
```diff
@@ -271,13 +269,10 @@ func TestMyLogIngester_ProcessLog(t *testing.T) {
...
-     Severity:  "INFO",
-    },
-    &googlecloudcommon_contract.GCPMainMessageFieldSet{
-     MainMessage: "server started",
+   input: testlog.NewMockLog(
+    time.Date(2026, 5, 22, 12, 0, 0, 0, time.UTC),
+    map[string]any{
+     "message": "server started",
     },
    ),
```
kyasbal: This looks not valid. MockLog can receive multiple fieldsets for testing. There are several occurrances in this PR

.agents/skills/khi_parser/SKILL.md
kyasbal, line 159: And this is not valid for overriding field set from testing.

docs/en/task-system/02-log-processing-cookbook.md
```diff
@@ -180,34 +187,42 @@ type MyMapper struct {
...
+        Type: inspectioncore_contract.TimelineTypeK8sCluster,
+    })
+    nsPath := builder.TimelineAccumulator.GetPath(clusterPath, khifilev6.PathSegment{
+        Name: namespace,
+        Type: inspectioncore_contract.TimelineTypeNamespace,
+    })
+    return builder.TimelineAccumulator.GetPath(nsPath, khifilev6.PathSegment{
+        Name: podName,
+        Type: inspectioncore_contract.TimelineTypeResource,
+    })
```
kyasbal: This is not super wrong, but if MustXXXTImeline is defined in the other place, it's strongly recommended to use them to create parent timeline.
Otherwise we would need to fix a lot of places when we decided to change timeline hierarchical structures.

pkg/common/structured/reader.go
```diff
@@ -44,115 +48,230 @@ func NewNodeReader(node Node) *NodeReader {
...
+
+// Segments returns the underlying path segments.
+func (p FieldPath) Segments() []string {
+	return p.segments
+}
+
+// IsEmpty returns true if the field path is empty.
+func (p FieldPath) IsEmpty() bool {
+	return len(p.segments) == 0
+}
```
kyasbal: FieldPath struct is defined in the middle of other struct. This should be in different file.

.agents/skills/khi_parser/SKILL.md
```diff
@@ -145,29 +141,14 @@ type CustomAppFieldSet struct {
...
-  RequestID: reader.ReadStringOrDefault("request_id", ""),
-  Payload:   reader.ReadStringOrDefault("payload", ""),
+// ExtractCustomApp extracts CustomAppFieldSet from a raw log node reader.
+func ExtractCustomApp(reader *structured.NodeReader) (CustomAppFieldSet, error) {
+ return CustomAppFieldSet{
+  AppName:   reader.ReadStringOrDefaultByPath(pathAppName, "unknown-app"),
+  RequestID: reader.ReadStringOrDefaultByPath(pathRequestID, ""),
+  Payload:   reader.ReadStringOrDefaultByPath(pathPayload, ""),
  }, nil
 }
```
kyasbal, line 164: In k8s audit log extractor, its input can be from OSS JSONL file or GCP cloud logging. Thus the extractor is injected in the task graph. This pattern should also be documented.

.agents/skills/khi_parser/SKILL.md
kyasbal: Done. Added notes in the skill documentation clarifying that pre-compiled FieldPath variables must be declared as package-level variables in a var block immediately below the imports.

.agents/skills/khi_parser/SKILL.md
kyasbal: Done. Updated the skill documentation to specify that ExtractXXX functions must return value types rather than pointer types to prevent heap allocation pressure.

.agents/skills/khi_parser/SKILL.md
kyasbal: Done. Updated the comment in the ProcessLog example to clarify that while l.Timestamp is usually used, custom timestamp fields can be extracted and set when higher precision is needed.

.agents/skills/log_timeline_mapper/SKILL.md
kyasbal: Done. Updated the testlog.NewMockLog testing example in the skill to use typed fieldsets instead of untyped maps and noted that multiple fieldsets can be passed.

.agents/skills/khi_parser/SKILL.md
kyasbal: Done. Added structured.GetMock check to the ExtractCustomApp example for unit test overriding.

docs/en/task-system/02-log-processing-cookbook.md
kyasbal: Done. Updated the cookbook documentation (in both English and Japanese) to demonstrate composing parent MustXXXTimeline helper functions rather than rebuilding parent segments from scratch.

pkg/common/structured/reader.go
kyasbal: Done. Moved FieldPath, CompileFieldPath, EmptyFieldPath, and ParseFieldPath to a dedicated fieldpath.go file and added fieldpath_test.go.

.agents/skills/khi_parser/SKILL.md
kyasbal: Done. Documented the Injected Extractor pattern (Pattern 2) in the khi_parser skill for multi-source log ingestion.

Issue #1018, opened 2026-09-17
[Epic] Timeline Time Range Filtering (Backend Pipeline & Interactive Frontend UI)


kyasbal:
# [Epic] Timeline Time Range Filtering (Backend Pipeline & Interactive Frontend UI)

## Background & Motivation
Currently, KHI's timeline and log filtering pipeline filters resources by kind, namespace, name regex, or CEL expressions across the entire dataset duration. When investigating incidents in large clusters, users frequently identify a critical log timestamp first and want to isolate only the resources and logs that experienced state changes (revisions or events) within a specific time window around that timestamp.

This epic introduces end-to-end **Time Range Filtering** across the Go backend workbench pipeline and the Angular frontend timeline view and toolbar.

## Core Specifications
1. **Backend Time Range Filter Pipeline**:
   - Extend `FilterTimelineRequest` and `FilterTimelineSyncRequest` protobuf messages to accept `filter_start_time` and `filter_end_time`.
   - When unspecified, the filter covers the full dataset duration (including resources created outside the query interval).
   - Index log timestamps on `cel.LogData` and `cel.EventInfo`, ensuring both `Events` and `Revisions` (`changedTime`) within each `TimelineData` are sorted chronologically.
   - Evaluate resource activity in $O(\log E + \log R)$ per timeline using binary search on sorted events and revisions.
   - **Hierarchy & Log Semantics**:
     - Time range filtering takes precedence over descendant expansion: child resources without any event or revision change within `[filter_start_time, filter_end_time]` are excluded, while parent/ancestor timelines are preserved via `IncludeAncestorsFilter` to maintain tree hierarchy.
     - Matched `LogIDs` are strictly restricted to logs whose timestamps fall within `[filter_start_time, filter_end_time]`.

2. **Frontend Interactive Selection (Ruler Drag Selection)**:
   - Dragging horizontally on the timeline ruler selects a time range directly, showing a dashed preview band while the pointer is held down.
   - Dragging right to left is normalized so the selection always has `startMs <= endMs`, and a drag shorter than 3px is treated as a click instead of a selection.
   - Releasing the mouse button confirms and applies the time filter; pressing `Escape` during the drag cancels the selection.
   - The active range can be adjusted afterwards by dragging its left or right edge handle, and cleared with the button rendered at the top of the range.

3. **Timeline Overlays & Toolbar Filter Chip**:
[...]

PR #1030, opened 2026-09-17
feat(frontend): add edge resize handles and clear button to the active time range on the ruler

kyasbal:
Part of #1018

Makes an applied time range filter directly editable on the timeline ruler.

### Summary of Changes

- **Edge resizing**
  - Added `TimeRangeDragMode` (`Create` / `ResizeStart` / `ResizeEnd`) to `interaction-model.ts`.
  - Rendered 8px resize handles on both edges of the active range overlay with an `ew-resize` cursor. Dragging a handle moves that edge while the opposite edge acts as the fixed anchor, and the range is normalized when an edge is dragged past the anchor.
  - During an edge resize the active overlay stays mounted but is dimmed through an `is-resizing` class, so the dashed preview shows the pending range while the handles keep the resize cursor. `Escape` cancels the resize and restores the previous range.
  - Extracted the shared drag setup into `startDrag` and replaced `isDragging` / `dragStartMs` with `activeDragMode` / `dragAnchorMs`.
  - Cached the ruler container's left offset per gesture in `cacheDragContainerLeft` so `mousemove` no longer forces a synchronous layout.
- **Clear button**
  - Added a small clear button at the top center of the active range overlay. Clicking it emits the new `timeRangeCleared` output, which `TimelineFrameComponent` propagates to `TimelineSmartComponent` to reset `ViewStateService.timeRangeFilter`.
- **Tests and Storybook**
  - Added unit tests for start/end edge resizing, boundary inversion, `Escape` cancellation, overlay dimming during resize, clear button emission, and the propagation path through the frame and smart components.
  - Added a `WithActiveTimeRange` story so the handles and the clear button are visible in Storybook.


kyasbal:
Thanks for the review! I applied the `is-resizing` approach across all 6 comments.

Instead of unmounting the active range overlay while an edge is dragged, the overlay now stays in the DOM and is dimmed via a CSS class, so the resize handles keep the `ew-resize` cursor for the whole gesture and no layout thrashing happens.

One deviation from the suggested diff: rather than widening `activeDragMode` to `protected`, I added a dedicated `protected readonly isResizingActiveRange` computed signal and bound `[class.is-resizing]="isResizingActiveRange()"`. This keeps the raw drag state private and, more importantly, keeps the class off during a `TimeRangeDragMode.Create` drag — `activeDragMode() !== null` would also have dimmed the existing range while the user drags out a brand new one on the ruler, which is not intended.

The SCSS now transparentizes the background and borders and hides the clear button under `.is-resizing`, and both affected unit tests assert the overlay stays mounted with `.is-resizing` during the drag and loses the class after mouseup / Escape. All 1400 frontend tests pass.

PR #1031, opened 2026-09-18
feat(frontend): render the time range filter chip label in two lines and add a detailed tooltip

kyasbal:
Part of #1018

Fixes the toolbar time range chip, whose label was always truncated at the end because the full `YYYY-MM-DD HH:mm:ss ~ YYYY-MM-DD HH:mm:ss` text did not fit in the chip.

### Summary of Changes

- **Two-line chip label**
  - Replaced `formatTimeRangeChipLabel` with `formatTimeRangeChipLines`, which returns a `TimeRangeChipLines` object with `startLine` and `endLine`. The start timestamp is rendered on the first line and `~ <end>` on the second line, so the chip stays narrow without any ellipsis. The date is repeated on the second line only when the range crosses a day boundary.
  - Laid out `.time-range-value` as a two-row CSS grid with a smaller font and tighter line height in both `ToolbarComponent` and `ToolbarAdvancedComponent`, and dropped the now meaningless `:` separator.
- **Detailed tooltip**
  - Added `formatTimeRangeTooltip`, which renders full ISO 8601 timestamps plus the duration, for example `2026-09-17T14:00:00.000+09:00 ~ 2026-09-17T14:30:00.000+09:00 (30m00s)`. The chip tooltip now shows this instead of a static string.
- **Duration formatting**
  - Extended `formatDurationMs` to hours and days (`1h00m`, `11h05m`, `2d03h`). Durations shorter than one hour keep their previous output. The existing expectation for exactly one hour changed from `60m00s` to `1h00m`, which is the intended behavior change.
- **Tests**
  - Covered same-day and cross-day ranges, non-zero timezone shifts, the non-positive timestamp fallback, and the new duration units, plus two-line DOM rendering and tooltip content in both toolbar components.


kyasbal:
Thanks for the review! I addressed one of the two points and intentionally skipped the other.

**Applied — rounding consistency.** You are right that the duration truncated sub-millisecond remainders while `formatIsoTimestampNs` rounds them, so the tooltip could show a duration that does not match the difference of the two timestamps it displays. Rather than duplicating the rounding expression, I extracted the conversion that was already inlined in `formatIsoTimestampNs` into a module-private `roundNsToMs(timestampNs: bigint): number` helper and reused it in `formatIsoTimestampNs`, `formatTimeRangeChipLines`, and `formatTimeRangeTooltip`. Now the chip, the tooltip timestamps, and the duration all agree on a single rounding rule. Added a regression test where the start rounds down and the end rounds up, which the previous truncating code would have reported 1ms short.

**Skipped — `Number.isFinite(timezoneShiftHours)` sanitization.** KHI's development rules prohibit defensive checks on required parameters unless the value has a distinct domain meaning. `timezoneShiftHours` is supplied by the `timezoneShift()` signal, which is always a finite number, so a `NaN` here would be a contract violation that should fail loudly rather than be silently coerced to UTC. Neither `formatIsoTimestampNs` nor `formatTimeRangeChipLines` sanitizes it either, so adding it in only one of the three formatters would also be inconsistent.

All 1409 frontend tests pass.

